PR #10, opened 2023-11-09
Fix regex issue and typos in Dockerfile rule

Edaaltuntas:
Hello there! 

While testing the `Dockerfile no latest tag` rule, I ran into a little hiccup. It turns out the rule wasn't quite doing what we expected. After a bit of sleuthing, I discovered that removing the caret symbol at the end of the regex pattern did the trick. I also took a moment to tidy up some older comments in the code.

Thank you for providing us with this wonderful project! Cheers 🍻

### DIFF
--- a/rule-types/github/dockerfile_no_latest_tag.yaml
+++ b/rule-types/github/dockerfile_no_latest_tag.yaml
@@ -22,12 +22,12 @@ def:
     git:
       branch: main
   # Defines the configuration for evaluating data ingested against the given profile
-  # This example uses the checks for that GitHub actions are using pinned tags
-  # for the use directive, in the form of SHA-1 hash.
-  # For example, this wil fail:
-  # uses: actions/checkout@v2
-  # This will pass:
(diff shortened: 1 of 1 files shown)


JAORMX (approved):

rule-types/github/dockerfile_no_latest_tag.yaml
```diff
@@ -40,7 +40,7 @@ def:
           dockerfile := file.read("Dockerfile")
 
           # Find all lines that start with FROM
-          from_lines := regex.find_n("^FROM \\S+.*^", dockerfile, -1)
+          from_lines := regex.find_n("^FROM \\S+.*", dockerfile, -1)
```
JAORMX, line 43: Let me give this one some thought. My only concern is that this will miss multi-stage builds. We want to match until the newline or a whitespace.

JAORMX: That works for me. Thanks for looking into this!

PR #11, opened 2023-11-14
Skip secret_push_protection and secret_scanning for private repos

jhrozek:
We might want to have a different pair of rules for github Pro users,
but I think it's mroe pressing to make the existing rules work for our
setup and having the rules seamlessly skip private repos would allow us
to use more rules across more repositories in the stacklok org.

### DIFF
--- a/rule-types/github/secret_push_protection.yaml
+++ b/rule-types/github/secret_push_protection.yaml
@@ -4,7 +4,12 @@ type: rule-type
 name: secret_push_protection
 context:
   provider: github
-description: Verifies that secret push protection is enabled for a given repository.
+description: |
+  Verifies that secret push protection is enabled for a given repository.
+  Note that this will will not work as expected for private repositories
+  unless you have GitHub Advanced Security enabled. If you still want to use
+  this rule because you have a mixture of private and public repositories,
+  enable the `skip_private_repos` flag.
 guidance: |
   You can use secret scanning to prevent supported secrets from being pushed into your repository by enabling secret scanning push protection.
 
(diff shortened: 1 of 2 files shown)


rule-types/github/secret_push_protection.yaml
```diff
@@ -34,14 +34,24 @@ def:
...
+
+        default allow := false
+        default skip := false
+
+        allow if {
+          input.ingested.security_and_analysis.secret_scanning_push_protection.status == "enabled"
+        }
+
+        skip if {
+          input.ingested.private == true
```
JAORMX, line 53: Why not make this configurable?

JAORMX (approved):

rule-types/github/secret_push_protection.yaml
```diff
@@ -21,6 +21,11 @@ def:
       enabled:
         type: boolean
         default: true
+      skip_private_repos:
+        type: boolean
+        default: true
+        description: |
+          If true, this rule will be marked as skipped for private repositories
```
JAORMX, line 28: Can you add extra documentation here of why this would be useful? e.g. in the overall rule type description:

```
Verifies that secret push protection is enabled for a given repository. Note that this is a paid feature by GitHub which does not work for private repositories. If you want to skip this rule for private repositories, set the `skip_private_repos` flag.
```

rule-types/github/secret_scanning.yaml
```diff
@@ -23,6 +23,11 @@ def:
       enabled:
         type: boolean
         default: true
+      skip_private_repos:
+        type: boolean
+        default: true
+        description: |
+          If true, this rule will be marked as skipped for private repositories
```
JAORMX, line 30: Same as above, can you add extra explanation to the overall rule description?

JAORMX (changes requested):

JAORMX (approved):

PR #13, opened 2023-11-15
Add new rule-type that checks if automated branch deletion is enabled

JAORMX:
This rule checks whether automatic branch deletion is enabled or not [1].

This also adds the rule to our reference profiles.

[1] https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-the-automatic-deletion-of-branches

### DIFF
--- a/profiles/github/stacklok-profile-remediate.yaml
+++ b/profiles/github/stacklok-profile-remediate.yaml
@@ -8,6 +8,9 @@ context:
 alert: "off"
 remediate: "on"
 repository:
+  - type: automatic_branch_deletion
+    def:
+      enabled: true
   - type: secret_scanning
     def:
       enabled: true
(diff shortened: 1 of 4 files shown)


profiles/github/stacklok-profile-remediate.yaml
```diff
@@ -8,6 +8,9 @@ context:
 alert: "off"
 remediate: "on"
 repository:
+  - type: automatic_branch_deletion
```
jhrozek, line 11: we also have stacklok-profile-read-only, can you add the rule there, too?

JAORMX: sure! will do.

JAORMX: Done!

jhrozek (approved):

PR #18, opened 2023-11-16
Don't require linear history

jhrozek:
Different stacklok repositories have different opinions, so let's no
enable this rule.


JAORMX:
Why? Can't we just converge?

jhrozek:
> Why? Can't we just converge?

I'll ask around but the last time this was a contentious topic and required a poll and show of hands ;-)

JAORMX (approved):

rdimitrov (approved):

PR #20, opened 2023-11-24
Add a new rule_type that allows to check the actions in a repo against a set of allowed actions

jhrozek:
Adds a new rule_type that allows specifying an allow list of actions
used in repositories. The evaluation uses rego and is able to print a
JSON representation of the actions that are not allowed.

Example profile:
```
version: v1
type: profile
name: actions-github-profile
context:
  provider: github
alert: "off"
remediate: "off"
repository:
  - type: repo_action_list
    def:
      actions:
        - actions/checkout
        - docker/build-push-action
        - docker/login-action
        - docker/metadata-action
        - docker/setup-buildx-action
        - sigstore/cosign-installer
```

Specifying an empty list allows you to list the actions used in the repo:
```
version: v1
type: profile
name: actions-github-profile
context:
  provider: github
alert: "off"
remediate: "off"
repository:
  - type: repo_action_list
    def:
      actions: []
```

Note that this PR depends on https://github.com/stacklok/minder/pull/1728

### DIFF
--- a/rule-types/github/repo_action_list.yaml
+++ b/rule-types/github/repo_action_list.yaml
@@ -1,0 +1,64 @@
+---
+version: v1
+type: rule-type
+name: repo_action_allow_list
+context:
+  provider: github
+description: |
+  Verifies that the github workflows in a repo only use actions enumerated in the rule.
+guidance: |
+  Having an overview over which actions and reusable workflows are allowed in a repository is important and allows for a better overall security posture.
+
+  For more information, see
+  https://docs.github.com/en/rest/actions/permissions#set-allowed-actions-and-reusable-workflows-for-a-repository
+def:
+  in_entity: repository
+  rule_schema:
+    type: object
+    properties:
+      actions:
+        type: array
+        description: |
+          List of actions that are allowed to be used in the repository.
+          The list should be in the format of `owner/repo`
+          For example, `actions/checkout`
+        items:
+            type: string
+  ingest:
+    type: git
+    git:
+      branch: main
+  eval:
+    type: rego
+    rego:
+      type: constraints
+      violation_format: json
+      def: |
+        package minder
+
+        expected_set := {x | x := input.profile.actions[_]}
+        actions := github_workflow.ls_actions("./.github/workflows")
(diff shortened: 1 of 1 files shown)


jhrozek:
moved out of draft as the rule seems to work well

JAORMX (approved):

rdimitrov (approved):

rule-types/github/repo_action_list.yaml
```diff
@@ -0,0 +1,64 @@
...
+        package minder
+
+        expected_set := {x | x := input.profile.actions[_]}
+        actions := github_workflow.ls_actions("./.github/workflows")
+
+        violations[{"msg": msg}] {
+          intersection := expected_set & actions
+          not count(intersection) == count(actions)
+
+          difference := [x | x := actions[_]; not intersection[x]]
```
evankanderson, line 46: Can this simply be `expected_set - actions`?  If so, you could rewrite this as:

```suggestion
          missing := expected_set - actions
          not count(missing) == 0
          msg = format_message(missing, input.output_format)
```

evankanderson (approved):
I think I understand what this is doing, though rego branching feels a bit weird to me.

rule-types/github/repo_action_list.yaml
jhrozek: I think we could use a combination of the two. The reason I used the weird list comprehension that creates the `difference` list was that IIRC I had issues formatting a set type later in the formatter. I think the difference you used reads better than the intersection, but I will also test edge cases (empty expected_set or expected_set that contains fewer items than actions). I'd like to use your proposal because it reads better, but I'm not that familiar with rego.

JAORMX: What if we merge the implementation as-is and push @evankanderson 's fix in a subsequent PR?

jhrozek: in my local testing a variant of what Evan suggested seems to work well. I'll give it a bit to test in staging

JAORMX (approved):

PR #23, opened 2023-12-07
Add rule type to verify open security advisories

JAORMX:
This verifies that security advisories have been triaged and either
closed or published. The intent is that projects should take their advisories
serisouly and make an active effort in keeping them in check.


profiles/github/profile.yaml
```diff
@@ -5,7 +5,7 @@ type: profile
 name: acme-github-profile
 context:
   provider: github
-alert: "on"
+alert: "off"
```
rdimitrov, line 8: Is this intended?

rdimitrov:
> @rdimitrov I added a threshold parameter. I was unable to add the date check within a reasonable amount of time. Can we do this in a separate PR?

Yeah, just create an issue once we merge this 👍

JAORMX:
I thought I had created an issue about this in the morning when I found it.... Let me find it.

PR #29, opened 2024-01-16
Add remediation to rule that checks trivy

JAORMX:
This creates a dedicated workflow that will run trivy on a repository.

### DIFF
--- a/rule-types/github/trivy_action_enabled.yaml
+++ b/rule-types/github/trivy_action_enabled.yaml
@@ -71,3 +71,34 @@ def:
     type: security_advisory
     security_advisory:
       severity: "medium"
+  remediate:
+    type: pull_request
+    pull_request:
+      title: "Add Trivy repository scan configuration"
+      body: |
+        This is a Minder automated pull request.
+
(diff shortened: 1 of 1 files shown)


rule-types/github/trivy_action_enabled.yaml
```diff
@@ -70,4 +70,35 @@ def:
...
\ No newline at end of file
+      severity: "medium"
+  remediate:
+    type: pull_request
+    pull_request:
+      title: "Add Trivy repository scan configuration"
+      body: |
+        This is a Minder automated pull request.
+
+        This pull request adds a GitHub actions workflow  that runs Trivy to the repository.
```
rdimitrov, line 81:
```suggestion
        This pull request adds a GitHub actions workflow that runs Trivy to the repository.
```

rdimitrov (dismissed):
Not related to this PR, but at some point we should address the problem of making sure the versions we suggest through remediations are up-to-date 👍

JAORMX:
@rdimitrov applied your suggestion

PR #31, opened 2024-01-16
Add profile with common github workflow security settings

JAORMX:


profiles/github/workflow_security.yaml
```diff
@@ -0,0 +1,18 @@
+---
+# A profile to verify common security settings for GitHub workflows
+version: v1
+type: profile
+name: workflow-security-github-profile
+context:
+  provider: github
+alert: "off"
+remediate: "off"
```
jhrozek, line 9: did you want these off by default?

JAORMX: uhm... good point, let me change it.

jhrozek (approved):

PR #35, opened 2024-01-17
Add new simple demo-able profiles

JAORMX:


jhrozek (approved):
@JAORMX showed me the profiles interactively, asking

PR #36, opened 2024-01-22
Update ghas.yaml example profile to include dependabot_configured rule type

meganbruce:
Added the dependabot_configured.yaml rule type to this example profile (for golang). I'm proposing we do this for completeness of the example profile and a docs page I'm working on to show how to use Minder with GHAS.


meganbruce:
@JAORMX would you mind reviewing (per our Slack convo)?

JAORMX (approved):

PR #39, opened 2024-01-24
Drop the v from excludes of slsa-github-generator

jhrozek:


jhrozek:
it's the other way around!

PR #41, opened 2024-01-24
Remove the stacklok read-only profile

jhrozek:
Maintaining two profiles was tedious.

Note: the remediating profile doesn't contain the trivy action and the
`allowed_selected_actions` which only allows actions based on whether
they are verified, github created or a pattern.


JAORMX (approved):

jhrozek:
rebased on main, no other changes

JAORMX (approved):

PR #48, opened 2024-02-07
Rename the workflow_name to signer_identity for artifact_signature

rdimitrov:
This renames the field so it's more generic since we'll allow in Minder to specify other types of identities, not just a workflow name.

It also adds 2 example profiles using the ruletype - a simple one and an extended version of it.

Related to: https://github.com/stacklok/minder/issues/2298


jhrozek (approved):
let's just merge this after https://github.com/stacklok/minder/pull/2299

PR #51, opened 2024-02-09
Add homoglyphs rule types

teodor-yanev:
The implications of homoglyphs type of attacks are described in https://github.com/stacklok/minder/issues/2121
It also contains additional explicatory links and useful examples.

### DIFF
--- a/rule-types/github/homoglyphs_invisible_characters.yaml
+++ b/rule-types/github/homoglyphs_invisible_characters.yaml
@@ -1,0 +1,23 @@
+version: v1
+type: rule-type
+name: invisible_characters_check
+context:
+  provider: github
+description: Verifies that pull requests do not add any invisible characters
+guidance: >
+  For every pull request submitted to a repository, this rule will check if the
+  pull request adds a new dependency with invisible characters. If it does, the rule will
+  fail and the pull request will be commented on.
+def:
+  in_entity: pull_request
+  param_schema:
+    properties: {}
+  rule_schema: {}
+  ingest:
+    type: diff
+    diff:
+      type: full
+  eval:
+    type: homoglyphs
+    homoglyphs:
+      type: invisible_characters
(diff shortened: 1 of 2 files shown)


jhrozek (commented):
could we add links that describe what kind of attack this prevents? I don't think they are very well known

rule-types/github/homoglyphs_invisible_characters.yaml
```diff
@@ -0,0 +1,23 @@
+version: v1
+type: rule-type
+name: invisible_characters_check
+context:
+  provider: github
+description: Verifies that pull requests do not add any invisible characters
+guidance: >
+  For every pull request submitted to a repository, this rule will check if the
+  pull request adds a new dependency with invisible characters. If it does, the rule will
+  fail and the pull request will be commented on.
```
JAORMX, line 10: The guidance is meant to tell you what to do if the rule fails. This is more appropriate for a description.

rule-types/github/homoglyphs_mixed_scripts.yaml
```diff
@@ -0,0 +1,23 @@
+version: v1
+type: rule-type
+name: mixed_scripts_check
+context:
+  provider: github
+description: Verifies that pull requests do not add any text with mixed scripts
+guidance: >
+  For every pull request submitted to a repository, this rule will check if the
+  pull request adds text with mixed scripts. If it does, the rule will
+  fail and the pull request will be commented on.
```
JAORMX, line 10: The guidance is meant to tell you what to do if the rule fails. This is more appropriate for a description.

JAORMX (changes requested):

rule-types/github/homoglyphs_mixed_scripts.yaml
teodor-yanev: Thanks: updated

JAORMX (approved):

PR #53, opened 2024-02-19
Enable and fix Go support for the Trusty and OSV rule types

rdimitrov:
The following PR:
* updates the `pr_trusty_check` rule so it now enables the go support in Trusty
* points the Trusty instance to the production one.
* updates the `pr_vulnerability_check` rule so it uses go.mod instead of go.sum for parsing

Example use:
* https://github.com/rdimitrov/bad-repo-go/pull/19

### DIFF
--- a/rule-types/github/pr_trusty_check.yaml
+++ b/rule-types/github/pr_trusty_check.yaml
@@ -44,7 +44,10 @@ def:
           depfile: package-lock.json
         - name: pypi
           depfile: requirements.txt
+        - name: go
+          depfile: go.mod
   # Defines the configuration for evaluating data ingested against the given profile
   eval:
     type: trusty
-    trusty: {}
+    trusty:
+      endpoint: "https://trusty.stacklok.dev"
--- a/rule-types/github/pr_trusty_check.yaml
+++ b/rule-types/github/pr_trusty_check.yaml
@@ -49,5 +49,3 @@ def:
   # Defines the configuration for evaluating data ingested against the given profile
   eval:
     type: trusty
-    trusty:
-      endpoint: "https://trusty.stacklok.dev"
--- a/rule-types/github/pr_vulnerability_check.yaml
+++ b/rule-types/github/pr_vulnerability_check.yaml
@@ -66,7 +66,7 @@ def:
         - name: npm
           depfile: package-lock.json
         - name: go
-          depfile: go.sum
+          depfile: go.mod
         - name: pypi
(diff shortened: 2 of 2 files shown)


rule-types/github/pr_trusty_check.yaml
```diff
@@ -44,7 +44,10 @@ def:
...
         - name: pypi
           depfile: requirements.txt
+        - name: go
+          depfile: go.mod
   # Defines the configuration for evaluating data ingested against the given profile
   eval:
     type: trusty
-    trusty: {}
+    trusty:
+      endpoint: "https://trusty.stacklok.dev"
```
jhrozek, line 53: Don't we want to let minder handle the defaults and deprecate this field?

rdimitrov: @jhrozek - What do you think about the following?
* Remove the trusty field entirely from the rule type - https://github.com/stacklok/minder/blob/c8ddfe603f7f78af75e0d2b093d0bd1c11ea03b0/proto/minder/v1/minder.proto#L1315
* Move reading the endpoint from here - https://github.com/stacklok/minder/blob/c8ddfe603f7f78af75e0d2b093d0bd1c11ea03b0/internal/engine/eval/eval.go#L59 to https://github.com/stacklok/minder/blob/c8ddfe603f7f78af75e0d2b093d0bd1c11ea03b0/internal/engine/eval/trusty/trusty.go#L45 where we'll read the env var and if missing default to `"https://trusty.stacklok.dev"`

JAORMX:
Is this in prod already?

rdimitrov:
> Is this in prod already?

Not yet from what I can see - prod uses `0.20240209.1326+ref.44dba51` and we merged the fix for this on Minder on Feb, 10th.

rule-types/github/pr_trusty_check.yaml
jhrozek: sounds good to me! Bonus point for adding debug messages to see what endpoint did we select and based on what.

rdimitrov: Of course, aiming for the jackpot 😃

rdimitrov:
Related: https://github.com/stacklok/minder/pull/2381

jhrozek (dismissed):
Works fine with the current main, but do not push until we bump prod

JAORMX (approved):

PR #55, opened 2024-02-27
Add profile for Stacklok health check

JAORMX:
This is a profile meant to verify the general overall health of an
organization in an easy way. It does not alert nor remediate by default.

Co-Authored-By: Radoslav Dimitrov <radoslav@stacklok.com>
Co-Authored-By: Teodor Yanev <teodor@stacklok.com>
Co-Authored-By: Eleftheria Stein-Kousathana <eleftheria@stacklok.com>
Co-Authored-By: Don Browne <dmjb@users.noreply.github.com>
Co-Authored-by: Adolfo Garcia Veytia (puerco) <puerco@stacklok.com>
Co-Authored-by: Evan Anderson <evan@stacklok.com>
Co-Authored-by: Brian Dussault <dussaultbrian@gmail.com>

### DIFF
--- a/profiles/github/stacklok-health-check.yaml
+++ b/profiles/github/stacklok-health-check.yaml
@@ -11,6 +11,9 @@ repository:
   - type: actions_check_pinned_tags
     def:
       exclude:
+        # generator_generic_slsa3 does not support pinning and will fail to retrieve the
+        # generator binary. We need to exclude it from pinning because of this.
+        # See https://github.com/slsa-framework/slsa-github-generator/issues/2993
         - slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@v1.9.0
   - type: branch_protection_allow_force_pushes
     def: 
--- a/profiles/github/stacklok-health-check.yaml
+++ b/profiles/github/stacklok-health-check.yaml
@@ -30,16 +30,19 @@ repository:
     name: go_dependabot
     def:
       package_ecosystem: gomod
+      schedule_interval: ""
       apply_if_file: go.mod
   - type: dependabot_configured
     name: npm_dependabot
     def:
       package_ecosystem: npm
+      schedule_interval: ""
       apply_if_file: package.json
   - type: dependabot_configured
     name: pypi_dependabot
     def:
       package_ecosystem: pypi
+      schedule_interval: ""
       apply_if_file: requirements.txt
   - type: dockerfile_no_latest_tag
     def: {}
@@ -54,5 +57,7 @@ repository:
 pull_request:
   - type: invisible_characters_check
     def: {}
+    params: {}
   - type: mixed_scripts_check
     def: {}
+    params: {}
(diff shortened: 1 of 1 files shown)


profiles/github/stacklok-health-check.yaml
```diff
@@ -0,0 +1,66 @@
...
+# Stacklok health check profile.
+version: v1
+type: profile
+name: stacklok-health-check
+context:
+  provider: github
+alert: "off"
+remediate: "off"
+repository:
+  # IMPORTANT: This needs to be paired with a rule
```
jhrozek, line 11: not related to this PR per se, but it would be cool if we had a "enable dependabot and guess the package ecosystems" rule. I wonder if we could use the dependency graph API to get the ecosystems or just write a simple "find" to set up the dependencies - but that would then require a special dependabot remediator which would create a PR based on the git contents or pass the found ecosystems from the git ingestor to the remediator.

profiles/github/stacklok-health-check.yaml
```diff
@@ -0,0 +1,66 @@
...
+    def: {}
+  - type: secret_push_protection
+    def:
+      enabled: true
+      skip_private_repos: true
+  - type: "secret_scanning"
+    def:
+      enabled: true
+      skip_private_repos: true
+# TODO: Should we enable this?
```
jhrozek, line 59: Since most of the rules here would fail (the branch protections for example) I don't see a huge deal in failing the artifact signatures as well.

profiles/github/stacklok-health-check.yaml
```diff
@@ -0,0 +1,66 @@
...
+# artifact:
+#  - type: artifact_signature
+#    params:
+#      sigstore: tuf-repo-cdn.sigstore.dev
```
jhrozek, line 63: This is the default.

profiles/github/stacklok-health-check.yaml
jhrozek: discussed today

profiles/github/stacklok-health-check.yaml
jhrozek: this rule was removed

profiles/github/stacklok-health-check.yaml
jhrozek: this rule was removed

profiles/github/stacklok-health-check.yaml
```diff
@@ -0,0 +1,58 @@
...
+name: stacklok-health-check
+context:
+  provider: github
+alert: "off"
+remediate: "off"
+repository:
+  - type: actions_check_pinned_tags
+    def:
+      exclude:
+        - slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@v1.9.0
```
ethomson, line 14: I don't think that I understand this exclusion? 🤔 

Apologies if we talked about it and I am forgetting the reasoning.

JAORMX: I added a comment to explain this.

ethomson: Ah, yes, this does ring a bell. Appreciate the link to the discussion, that's very interesting.

ethomson (approved):

dmjb (approved):

PR #61, opened 2024-03-06
Fix codeQL remediation

jhrozek:
For some reason we were only setting up codeQL, but not actually running
it.


rule-types/github/codeql_enabled.yaml
```diff
@@ -66,6 +66,7 @@ def:
             # Ensure a workflow contains the codel-ql action
             some i
             steps := workflow.jobs.analyze.steps[i]
+            print(steps)
```
JAORMX, line 69: let's remove this `print`

JAORMX (changes requested):

rule-types/github/codeql_enabled.yaml
jhrozek: oops, leftover debugging from where I thought we might have a bug in the rego code :-)

JAORMX (approved):

PR #65, opened 2024-03-13
Add rule types severity

teodor-yanev:
Fixes https://github.com/stacklok/minder-rules-and-profiles/issues/64

### DIFF
--- a/rule-types/github/invisible_characters_check.yaml
+++ b/rule-types/github/invisible_characters_check.yaml
@@ -1,6 +1,8 @@
 version: v1
 type: rule-type
 name: invisible_characters_check
+severity:
+  value: medium
 context:
   provider: github
 description: |  
@@ -29,3 +31,6 @@ def:
     type: homoglyphs
     homoglyphs:
       type: invisible_characters
+  alert:
+    type: security_advisory
+    security_advisory: {}
--- a/rule-types/github/invisible_characters_check.yaml
+++ b/rule-types/github/invisible_characters_check.yaml
@@ -2,7 +2,7 @@ version: v1
 type: rule-type
 name: invisible_characters_check
 severity:
-  value: medium
+  value: high
 context:
   provider: github
 description: |  
--- a/rule-types/github/license.yaml
+++ b/rule-types/github/license.yaml
@@ -2,6 +2,8 @@
 version: v1
 type: rule-type
 name: license
+severity:
+  value: medium
 context:
   provider: github
 description: Verifies that there's a license file of a given type present in the repository.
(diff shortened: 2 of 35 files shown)


JAORMX:
Have you tested if upgrade of these rules work? I haven´t tried it out, but it would be worth doing it.

In a local env, do `minder ruletype create` with the rules before this change. Then do a `minder ruletype apply` with the change, and see that it all goes smoothly and the severity is indeed updated.

rule-types/github/invisible_characters_check.yaml
```diff
@@ -1,6 +1,8 @@
 version: v1
 type: rule-type
 name: invisible_characters_check
+severity:
+  value: medium
```
JAORMX, line 5: High? Maybe

rule-types/github/license.yaml
```diff
@@ -2,6 +2,8 @@
 version: v1
 type: rule-type
 name: license
+severity:
+  value: medium
```
JAORMX, line 6: Low

rule-types/github/mixed_scripts_check.yaml
```diff
@@ -1,6 +1,8 @@
 version: v1
 type: rule-type
 name: mixed_scripts_check
+severity:
+  value: medium
```
JAORMX, line 5: High? Maybe

rule-types/github/no_open_security_advisories.yaml
```diff
@@ -2,6 +2,8 @@
 version: v1
 type: rule-type
 name: no_open_security_advisories
+severity:
+  value: medium
```
JAORMX, line 6: Low

rule-types/github/repo_action_allow_list.yaml
```diff
@@ -2,6 +2,8 @@
 version: v1
 type: rule-type
 name: repo_action_allow_list
+severity:
+  value: medium
```
JAORMX, line 6: Info

rule-types/github/secret_push_protection.yaml
```diff
@@ -2,6 +2,8 @@
 version: v1
 type: rule-type
 name: secret_push_protection
+severity:
+  value: medium
```
JAORMX, line 6: High

rule-types/github/secret_scanning.yaml
```diff
@@ -2,6 +2,8 @@
 version: v1
 type: rule-type
 name: secret_scanning
+severity:
+  value: medium
```
JAORMX, line 6: High

teodor-yanev:
@JAORMX Thanks for your feedback! I updated the rules.

PR #71, opened 2024-04-01
Add a workflow that checks if all existing ruletypes have tests

rdimitrov:
Fixes https://github.com/stacklok/minder-smoke-tests/issues/84


puerco:
Let's rekick the job once [minder-smoke-tests#83](https://github.com/stacklok/minder-smoke-tests/pull/83) merges

puerco (approved):

PR #72, opened 2024-04-02
Update "no open security advisories rule"

teodor-yanev:
Please see https://stacklok.slack.com/archives/C057KR4DT0W/p1711998911156389 for details around the discussion. 

TL;DR: GitHub webhook events are not being sent when a draft GHSA is created. The two events related to it are when a GHSA is either published or reported. Reported is what the state of a GHSA is when in a "Triage" state (created).

Testing performed in my public testing repo: https://github.com/teodor-yanev/a-testrepo/security/advisories

"Triage" GHSAs can only be created when the 'Private vulnerability reporting' option is enabled, which is why I explicitly stated it in the description.

Also, to avoid confusing users, I added the additional "draft" state not to be considered when initially ingesting GHSAs. This only happens when the profile is created on a repo that already has a bunch of GHSAs.

### DIFF
--- a/rule-types/github/no_open_security_advisories.yaml
+++ b/rule-types/github/no_open_security_advisories.yaml
@@ -14,10 +14,14 @@ description: |
 guidance: |
   Ensuring that a repository has no open security advisories helps maintain a secure codebase.
 
-  This rule will fail if the repository has unacknowledged security advisories.
-  It will also fail if the repository has no security advisories enabled.
+  The rule will fail if:
+  - The repository has unacknowledged open security advisories in a "Triage" state.
+  - Security advisories are not enabled for the repository.
+  
+  Note:
+  Advisories in a "Triage" state are considered if the repository has enabled the 'Private vulnerability reporting' option.
 
-  Security advisories that are closed or published are considered to be acknowledged.
+  Security advisories that are draft, closed or published are considered to be acknowledged.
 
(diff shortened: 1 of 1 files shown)


rule-types/github/no_open_security_advisories.yaml
```diff
@@ -91,8 +95,7 @@ def:
         	# Is not withdrawn
         	adv.withdrawn_at == null
         
-        	adv.state != "closed"
-        	adv.state != "published"
+        	not adv.state in {"closed", "published", "draft"}
```
JAORMX, line 98: Wait, so this will no longer trigger on draft?

JAORMX (changes requested):

rule-types/github/no_open_security_advisories.yaml
rdimitrov: @JAORMX - I met with Teo and I think I understand why we want it that way:

1. Apparently GitHub doesn't do webhook events for opening security advisories with `state: draft`. 
* Reasoning is they considered them private and bla-bla - 
image

* Because of that it feels better to be consistent and not check for them at all since we don't have means to get events for opened `draft` advisories.

2. What we'll rely on instead is checking if there's something that is in `triage` state. 
* This means someone else (not the owner) reported something and it is still not processed (neither closed or published/acknowledged). 

WDYT?

JAORMX: that makes sense to me. Let's go for this

JAORMX (approved):

PR #83, opened 2024-04-10
Make guidance actionable in rule types

JAORMX:
The guidance is supposed to tell the user how to react when they find a
faliure or an issue in their profiles. This changes the verbiage of the
rule types guidances to reflect this. It should be a call to action
instead.


dmjb (approved):
Looks good, but we should copy this across to the healthcheck bundle repo too (we definitely need some better process around this)

PR #90, opened 2024-04-12
Minor ruletype descriptions

jhrozek:
A bunch of ruletype text updates I found while I was testing different
ruletypes through UI

- more ecosystem examples for the popular dependabot ecosystems
- provide a reasonable default for codeQL schedule
- say that minder will use builtin defaults for the OSV ruletype to make
  it clear that the user doesn't have to fill them in
- fix description of exclude for the action pinning ruletype

### DIFF
--- a/rule-types/github/codeql_enabled.yaml
+++ b/rule-types/github/codeql_enabled.yaml
@@ -38,7 +38,8 @@ def:
       schedule_interval:
         type: string
         description: |
-          Only applicable for remediation. Sets the schedule interval for the workflow.
+          Sets the schedule interval in cron format for the workflow. Only applicable for remediation.
+        default: '30 * * * *'
     required:
       - languages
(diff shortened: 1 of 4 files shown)


rule-types/github/codeql_enabled.yaml
```diff
@@ -38,7 +38,8 @@ def:
       schedule_interval:
         type: string
         description: |
-          Only applicable for remediation. Sets the schedule interval for the workflow.
+          Sets the schedule interval in cron format for the workflow. Only applicable for remediation.
+        default: '30 * * * *'
```
jhrozek, line 42: I /think/ this should be safe because we don't actually enforce the defaults on the back end level, so the defaults would be kind of advisory use for FE.

JAORMX (approved):

PR #91, opened 2024-04-12
Reorder Health Check profile

ethomson:


profiles/github/stacklok-health-check.yaml
```diff
@@ -9,6 +9,16 @@ context:
...
+  - type: secret_scanning
+    name: "Secret Scanning is enabled"
+    def:
+      enabled: true
+      skip_private_repos: true
+  - type: secret_push_protection
+    name: "Secret push protection is enabled"
+    def:
+      enabled: true
+      skip_private_repos: true
```
JAORMX, line 21: This duplicates these checks. Why is this reorder needed?

ethomson: Fixed. Wanted to move the higher priority ("more bang for the buck") rules higher in the priority list.

JAORMX (approved):

Issue #103, opened 2024-04-29
The artifact signature rule doesn't expose any details about what failed


jhrozek:
Because the artifact_signature rule uses the deny-by-default model all we get from the evaluation in case it fails is `denied`. Especially for artifacts this is not ideal as we can't tell the user what changed in the artifact vs. the expected state unless they look at minder logs.

PR #104, opened 2024-04-29
Change the artifact_signature rule to use constraints evaluation, not deny-by-default

jhrozek:
Because the artifact_signature rule uses the deny-by-default model all
we get from the evaluation in case it fails is denied. Especially for
artifacts this is not ideal as we can't tell the user what changed in
the artifact vs. the expected state unless they look at minder logs.

Fixes: #103

### DIFF
--- a/rule-types/github/artifact_signature.yaml
+++ b/rule-types/github/artifact_signature.yaml
@@ -85,21 +85,25 @@ def:
   eval:
     type: rego
     rego:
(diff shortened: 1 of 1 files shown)


JAORMX:
Let's brainstorm about this tomorrow.

JAORMX (approved):
I think this is fine given that we want to anyway refactor into smaller rule types in the future.

PR #105, opened 2024-05-02
Github SLSA Provenance

puerco:
Work in progress:

Attestation rules


jhrozek:
@puerco I pushed a new patch into your branch (hopefully that's OK since you can just delete it if you don't like the patch). I will point out the changes in inline comments.

rule-types/github/attestation_slsa_github.yaml
```diff
@@ -1,7 +1,7 @@
 ---
 version: v1
 type: rule-type
-name: artifact_signature
+name: artifact_signature_slsa
```
jhrozek, line 4: I think we should not change the existing type but add a new one.

rule-types/github/attestation_slsa_github.yaml
```diff
@@ -50,13 +50,15 @@ def:
       event:
         type: array
         description: "Events allowed to trigger a build"
-        enum:
-          - worflow_dispatch
-          - push
-        default: ["worflow_dispatch", "push"]
+        items:
+          type: string
```
jhrozek, line 54: I was getting errors unless I made the array type a string and then enumerated the allowed values.

rule-types/github/attestation_slsa_github.yaml
```diff
@@ -50,13 +50,15 @@ def:
...
+          enum:
+            - workflow_dispatch
+            - push
+        default: ["workflow_dispatch", "push"]
```
jhrozek, line 58: Note that the default has no effect on minder right now...

rule-types/github/attestation_slsa_github.yaml
JAORMX: I agree with this, if we want to change the signature, behavior or even deprecate the previous behavior, we should just document the deprecation and create a new rule type.

rule-types/github/attestation_slsa_github.yaml
```diff
@@ -91,43 +93,30 @@ def:
         import rego.v1
 
         default allow := false
+        default skip := false
         
-        allow if {
+        artifacts_github_provenance = {artifact |
```
jhrozek, line 98: this filter is the most important change and one I am not sure is correct. But based on my reading of your code what you wanted was to:
 - evaluate only artifacts whose `predicate_type` is `"https://slsa.dev/provenance/v1"` and the `buildType` is `"https://slsa-framework.github.io/github-actions-buildtypes/workflow/v1"` and skip the rest right?
 - those that match the two conditions above must match the profile

rule-types/github/attestation_slsa_github.yaml
```diff
@@ -91,43 +93,30 @@ def:
...
-        #  every artifactVersion in input.ingested {
-        #    if artifactVersion.Verification.attestation.predicateType == "https://slsa.dev/provenance/v1" {
-        #      artifactVersion.Verification["is_signed"]
-        #    }
-        ##    #every key, value in input.profile {
-        # #   #  artifactVersion.Verification[key] == value
-        #    #}
-        #  }
-        #}
+        skip if {
```
jhrozek, line 117: this should mark the rule as skipped if there are no matching artifacts (with "https://slsa-framework.github.io/github-actions-buildtypes/workflow/v1" == buildType and ""https://slsa.dev/provenance/v1" == predicate_type)

rule-types/github/attestation_slsa_github.yaml
```diff
@@ -91,43 +93,30 @@ def:
...
+          artifact.Verification.attestation.predicate.buildDefinition.buildType == "https://slsa-framework.github.io/github-actions-buildtypes/workflow/v1"
         }
 
+        allow if {
+          count(artifacts_github_provenance) > 0
+
+          every artifact in artifacts_github_provenance {
+            artifact.Verification.attestation.predicate.buildDefinition.externalParameters.workflow.path == input.profile.signer_identity
+            artifact.Verification.attestation.predicate.buildDefinition.externalParameters.workflow.ref == input.profile.workflow_ref
+            artifact.Verification.attestation.predicate.buildDefinition.externalParameters.workflow.repository == input.profile.workflow_repository
```
jhrozek, line 110: note that these conditions don't handle all the parameters in the `rule_schema` but I wasn't sure if you only want to match those under the attestation or also those in the top-level attributes.

JAORMX: Do we want to match all parameters? At some point it would be good just to have minimal rules that check different aspects of the attestation. That way we'd get more relevant error messages. Similar to what we did for branch protection rules.

jhrozek:
here's the profile I was testing with:
```
# sample policy for validating artifact signatures
version: v1
type: profile
name: artifact-signature-extended
context:
  provider: github
artifact:
  - type: artifact_signature_slsa
    params:
      tags: ["latest"]
      #tags: []
      name: good-repo-go
    def:
      is_signed: true
      is_verified: true
      workflow_repository: https://github.com/jakubtestorg/good-repo-go
      workflow_ref: refs/heads/main
      event: ["workflow_dispatch"]
      runner_environment: github-hosted
      signer_identity: .github/workflows/build-image-signed-ghat-malicious.yml
      cert_issuer: https://token.actions.githubusercontent.com # or for example - https://accounts.google.com
```

rule-types/github/attestation_slsa_github.yaml
puerco: Yes correct! Thank you :pray:

rule-types/github/attestation_slsa_github.yaml
puerco: Oh interesting. Is `skip` a  custom value that minder recognizes or is it just vanilla rego?

rule-types/github/attestation_slsa_github.yaml
puerco: yes totally agree i had not gotten to that part

rule-types/github/attestation_slsa_github.yaml
jhrozek: minder: https://github.com/stacklok/minder/blob/ee66f6c0763212503c898cfefb65ce1450c7f5ac/internal/engine/eval/rego/result.go#L94

rule-types/github/attestation_slsa_github.yaml
jhrozek: sorry, can you elaborate? Now we do match all defined as params in the ruletype did you mean all that exist in the ingestion result that we send?

jhrozek:
note: I just updated the `runner_environment` from simple string to a URL. I caught that when I was testing the profile for our blog post and fixed it there, but not here in the example in this PR.

puerco:
Closing this one in favor of #106 (from a branch, not a fork)

PR #111, opened 2024-05-09
Trusty PR add provenance and activity

puerco:
This PR updates the trusty PR rule to add the new sections for the score components which will be exposed to the evaluator once https://github.com/stacklok/minder/pull/3277 merges.

Note that `evaluate_score` isbeing removed because policies can now be written on all score components, not just one score.


stacklokbot (commented):
### :white_check_mark: No Invisible Unicode Characters Detected.

jhrozek (approved):

PR #112, opened 2024-05-10
Add dependabot profile for rust projects

blkt:
Add a new profile to configure dependabot check for Rust, cargo-based projects.


stacklokbot (commented):
### :white_check_mark: No Invisible Unicode Characters Detected.

JAORMX (approved):

PR #126, opened 2024-07-18
Add rule to detect GitHub Actions using default permissions

stacklokbot:
I built & presented this rule in a tutorial at CloudNativeSecurityCon: https://cloudnativesecurityconna24.sched.com/event/8470542653b26dbcab395ccc294b81b3

### DIFF
--- a/rule-types/github/actions_check_default_permissions.yaml
+++ b/rule-types/github/actions_check_default_permissions.yaml
@@ -1,0 +1,74 @@
+version: v1
+type: rule-type
+name: actions_check_default_permissions
+severity:
+  value: medium
+context:
+  provider: github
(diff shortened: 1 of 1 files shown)


jhrozek:
@evankanderson thanks for the PR. I think you opened the PR from the wrong account by mistake - do you want to fix it?
Also, we have a workflow that fails if there's no tests for a new ruletype. I can write one (I even think it's better if someone else than the author writes the test..), do I read the rego code correctly that all that would be needed would be a workflow that doesn't define permissions explicitly?

jhrozek:
tested manually, works fine.

evankanderson:
Replaced with #134

PR #131, opened 2024-07-26
Initial OpenSSF baseline profile

puerco:
This commit introduces the first draft of the security baseline profile. It introduces three simple ruletypes: Check for a security policy file, check  for a security insights file and check for a dependency policy in the Si file.

### DIFF
--- a/profiles/github/openssf_security_baseline.yaml
+++ b/profiles/github/openssf_security_baseline.yaml
@@ -13,8 +13,17 @@ repository:
   # (TODO) Memory Safety
   # (TODO) 2FA
   # (TODO) Secret scanning
-  # (TODO) Push protection
+  # Branch protection.
+  
+  # This enables general branch protection on the main branch, for
+  # other branch rules check the branch-protection.yaml profile 
+  - type: branch_protection_enabled
+    params:
+      branch: ""
+    def: {}
+
   # (TODO) Credentials with minimal permissions
+
   # Check for a Security Insights file in the repo
   - type: security_insights
     name: security_insights
--- a/profiles/github/openssf_security_baseline.yaml
+++ b/profiles/github/openssf_security_baseline.yaml
@@ -15,7 +15,13 @@ repository:
   # (TODO) Secret scanning
   # (TODO) Push protection
   # (TODO) Credentials with minimal permissions
-  # (TODO) Security Insights
+  # Check for a Security Insights file in the repo
+  - type: security_insights
+    name: security_insights
+    displayName: "Check for a Security Insights file in the repository"
+    def:
+      filename: SECURITY-INSIGHTS.yml
+
   # (TODO) Dependency Policy published
   # (TODO) Pinned dependencies
   # (TODO) Hardened Workflows
--- a/profiles/github/openssf_security_baseline.yaml
+++ b/profiles/github/openssf_security_baseline.yaml
@@ -1,0 +1,32 @@
+---
+# Profile to help comply with the Open Source Security Foundation's Security
+# Baseline, a framework of minimal security recommended to projects.
+version: v1
+type: profile
+name: openssf_security_baseline
+display_name: OpenSSF Security Baseline
+context:
+  provider: github
+alert: "on"
+remediate: "off"
+repository:
+  # (TODO) Memory Safety
+  # (TODO) 2FA
+  # (TODO) Secret scanning
+  # (TODO) Push protection
+  # (TODO) Credentials with minimal permissions
+  # (TODO) Security Insights
+  # (TODO) Dependency Policy published
+  # (TODO) Pinned dependencies
+  # (TODO) Hardened Workflows
+  # (TODO) Static code analysis
+  # (TODO) 2 person review
+  # (TODO) Security policy published
+  # (TODO) No vulnerabilities
+  # (TODO) 60 day SLA on HIGH/MED vulnerabilities
+  # (TODO) No checked-in binaries
+  # (TODO) SBOM generated with releases
+  # (TODO) SBOM conforms to NTIA minimal elements
+  # (TODO) Artifacts are signed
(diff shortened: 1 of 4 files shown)


rule-types/github/security_insights.yaml
```diff
@@ -0,0 +1,60 @@
...
+        type: string
+        description: |
+          The path to the Security Insights file
+        default: SECURITY-INSIGHTS.yml
+    required:
+      - filename
+  # Defines the configuration for ingesting data relevant for the rule
+  ingest:
+    type: git
+    git:
```
jhrozek, line 43: I think if you use an empty git configuration `git: {}` the ingestor would infer the default branch automatically IIRC. I agree we are not consistent about this in our own rules (we should!), but look at e.g. 42d619ecef3b431745f5999e4b0dd32c78cf4fc0

rule-types/github/security_insights.yaml
```diff
@@ -0,0 +1,60 @@
+---
+version: v1
+type: rule-type
+name: security_insights
```
jhrozek, line 4: This is a new rule and our CI will keep failing if we don't have either a test or an explicit skip placeholder. I recently added the placeholders for new rules in https://github.com/stacklok/minder-smoke-tests/commit/b69aadacb35dfd34221ae9d38cb71504cb8236a6 and https://github.com/stacklok/minder-smoke-tests/commit/4d89d22575a19575bf781ca27a9b7ab6539e38b4

profiles/github/openssf_security_baseline.yaml
```diff
@@ -13,8 +13,17 @@ repository:
...
   # (TODO) 2FA
   # (TODO) Secret scanning
-  # (TODO) Push protection
+  # Branch protection.
+  
+  # This enables general branch protection on the main branch, for
+  # other branch rules check the branch-protection.yaml profile 
+  - type: branch_protection_enabled
+    params:
+      branch: "main"
```
jhrozek, line 29: I think you can leave the branch empty to infer the default branch (We couldn't remove the branch param because it was marked as required)

profiles/github/openssf_security_baseline.yaml
```diff
@@ -35,7 +35,14 @@ repository:
...
   # (TODO) Hardened Workflows
   # (TODO) Static code analysis
-  # (TODO) 2 person review
+
+  # Force two-person review on pull requests before merging
+  - type: branch_protection_require_pull_request_approving_review_count
+    params:
+      branch: "main"
+    def:
+      required_approving_review_count: 2
```
jhrozek, line 69: Do they really require 2?

profiles/github/openssf_security_baseline.yaml
```diff
@@ -35,7 +35,14 @@ repository:
   # (TODO) Pinned dependencies
   # (TODO) Hardened Workflows
   # (TODO) Static code analysis
-  # (TODO) 2 person review
+
+  # Force two-person review on pull requests before merging
+  - type: branch_protection_require_pull_request_approving_review_count
+    params:
+      branch: "main"
```
jhrozek, line 67: Same comment about the hardcoded branch

profiles/github/openssf_security_baseline.yaml
```diff
@@ -41,7 +41,12 @@ repository:
   # (TODO) Dependency Policy published
   # (TODO) Pinned dependencies
   # (TODO) Hardened Workflows
-  # (TODO) Static code analysis
+
+  # Static code analysis: CodeQL
+  - type: codeql_enabled
+    def:
+      languages: [go, javascript, typescript]
```
jhrozek, line 61: It's fine if you want to hardcode the languages, but since ae92d747061a5ee9b0aee5e80b2f1bbec80da2aa you can also leave the list blank to let the rule (or rather the underlying action) autodetect the languages.

profiles/github/openssf_security_baseline.yaml
```diff
@@ -39,7 +39,19 @@ repository:
       filename: SECURITY-INSIGHTS.yml
 
   # (TODO) Dependency Policy published
-  # (TODO) Pinned dependencies
+
+  # Pinned dependencies
+  # This rule tells Minder to run Grizbee (https://github.com/stacklok/frizbee/)
```
jhrozek, line 45: Grizbee sounds like some energy drink with so much caffeine it's only sold on the black market :-)

jhrozek:
thanks for the work @puerco ! I left some comment inline - it's mostly that we developed some conventions over the months and they are not really visible for someone who's not been following the development. But overall this looks great!

profiles/github/openssf_security_baseline.yaml
puerco: Ah perfect, I was worried about this one.

profiles/github/openssf_security_baseline.yaml
puerco: Yes ;/

profiles/github/openssf_security_baseline.yaml
puerco: OK, done :+1:

rule-types/github/security_insights.yaml
puerco: OK, done. I've modified it on both rules :+1:

profiles/github/openssf_security_baseline.yaml
puerco: Fixed, but you have to admit it sounded cool :) 

image

rule-types/github/security_insights.yaml
puerco: OK, I opened this but ill write the tests as well
https://github.com/stacklok/minder-smoke-tests/pull/189

puerco:
Comments are addressed here, I'm closing this PR and reopening in #136 which is not from my fork

PR #141, opened 2024-08-22
Github: Add rule type that verifies that harden runner is the first step

JAORMX:
this adds a rule that verifies that the `harden-runner` action is part
of the first step of each workflow.

To test, you may call it as part of a profile that looks as follows:

```yaml
---
version: v1
type: profile
name: harden-runner
context:
  provider: github
repository:
  - type: step_security_harden_runner_enabled
    def: {}
```


rule-types/github/step_security_harden_runner_enabled.yaml
```diff
@@ -0,0 +1,71 @@
...
+  Set it up by adding the following as the first step of each job:
+
+  ```yaml
+    - uses: step-security/harden-runner@17d0e2bd7d51742c71671bd19fa12bdc9d40a3d6 # v2.8.1
+      with:
+        egress-policy: audit
+  ```
+  
+  For more information, see
+  https://github.com/marketplace/actions/aqua-security-trivy
```
jhrozek, line 25: Why do we point to trivy?

jhrozek:
The rule looks good, do you want to submit the placeholders for tests to get a green CI run?

jhrozek:
Crap, I clicked the wrong button and closed the PR by accident. Sorry!

rule-types/github/step_security_harden_runner_enabled.yaml
JAORMX: oops!!

JAORMX:
> The rule looks good, do you want to submit the placeholders for tests to get a green CI run?

@jhrozek I can do that. However, I don't think we should block on that. External contributors wouldn't be able to add new rule types.

jhrozek:
> > The rule looks good, do you want to submit the placeholders for tests to get a green CI run?
> 
> @jhrozek I can do that. However, I don't think we should block on that. External contributors wouldn't be able to add new rule types.

We don't have to block on that but should at least silence the warning by either writing a test or putting the placeholders in the test repo or else all subsequent PRs to this repo would fail this check.

jhrozek (approved):

evankanderson:
So, two issues that are a gap from where we are today:

# No remediations
This rule fires an alert, but despite the description of the remediation being pretty simple, we have no way to actually perform the remediation, because it involves:
1. Altering existing files, and we can only rewrite with mostly-fixed contents
2. We need to patch multiple arbitrary files, and we can only target a fixed number of fixed-name files.

# Incomplete checks and remediation
This only does [step 1 of the instructions](https://github.com/step-security/harden-runner?tab=readme-ov-file#hardening-github-hosted-runners).

> 1. Add the step-security/harden-runner GitHub Action to your GitHub Actions workflow file as the first step in each job. You can automate adding Harden-Runner Action to your workflow file by pasting your workflow in the StepSecurity online tool.
> 2. In the workflow logs and the job markdown summary, you will see a link to security insights and recommendations.
> 3. Click on the link ([example link](https://app.stepsecurity.io/github/step-security/github-actions-goat/actions/runs/7704454287)). You will see a process monitor view of network and file events correlated with each step of the job.
> 4. In the Recommended Policy tab, you'll find a recommended block policy based on outbound calls aggregated from the current and past runs of the job. You can update your workflow file with this policy, or alternatively, use the [Policy Store](https://docs.stepsecurity.io/harden-runner/how-tos/block-egress-traffic#2-add-the-policy-using-the-policy-store) to apply the policy without modifying the workflow file. From now on, any outbound calls not in the allowed list will be blocked.

I'm not sure how we'd do steps 3 & 4 to fill out the actual policy, assuming that Step Security provided an API for reading these recommendations.  We'd run into both the problems from "no remediations" as well as not having a way to fetch the output.

PR #146, opened 2024-08-23
Add a display name for each rule type

### LINKED ISSUE #144: Populate display names for all rule types

eleftherias:
Fix #144

### DIFF
--- a/rule-types/github/secret_scanning.yaml
+++ b/rule-types/github/secret_scanning.yaml
@@ -2,6 +2,7 @@
 version: v1
 type: rule-type
 name: secret_scanning
+display_name: Enable secret scanning to detect hardcoded secrets
 severity:
   value: high
 context:
--- a/rule-types/github/actions_check_default_permissions.yaml
+++ b/rule-types/github/actions_check_default_permissions.yaml
@@ -1,6 +1,7 @@
 version: v1
 type: rule-type
 name: actions_check_default_permissions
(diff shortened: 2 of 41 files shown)


JAORMX (approved):

JAORMX (dismissed):

.idea/vcs.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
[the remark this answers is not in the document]
jhrozek: Even better is to [use a global gitignore](https://stackoverflow.com/questions/7335420/global-git-ignore)

rule-types/github/secret_scanning.yaml
```diff
@@ -2,6 +2,7 @@
 version: v1
 type: rule-type
 name: secret_scanning
+display_name: Detect hardcoded secrets
```
ethomson, line 5: I know that this was explicitly the example that @jtenniswood used; but we don't do this. We ensure that secret scanning is enabled. Secret scanning does the detection. I don't _think_ that I'm just being pedantic here, as a user I would expect that "detect hardcoded secrets" would do just that and issue me _Minder alerts_ when I had secrets.

eleftherias: I see your point, I've updated the names for both "secret detection" rules.

rdimitrov (approved):

PR #160, opened 2024-09-18
Remove enabled property from ruletypes

### LINKED ISSUE #156: Remove enabled property from rule types
Certain rule types, like `secret_push_protection` and `secret_scanning` have a property called `enabled` which can be set when defining a profile.
We should remove that property and assume `enabled` is implicitly `true` when the user selects the given rule.

eleftherias:
Fix #156

### DIFF
--- a/profiles/github/branch-protection.yaml
+++ b/profiles/github/branch-protection.yaml
@@ -16,13 +16,11 @@ repository:
   - type: branch_protection_allow_deletions
     params:
       branch: ""
-    def:
-      allow_deletions: false
+    def: {}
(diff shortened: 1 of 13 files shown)


jhrozek:
we have also `.required_signatures.enabled` in `rule-types/github/branch_protection_require_signatures.yaml` although that does sound like something you might want to turn off explicitly. Same with `rule-types/github/branch_protection_enforce_admins.yaml` otherwise the changes LGTM (not reviewing since this is a draft)

jhrozek (approved):

PR #165, opened 2024-10-02
Use default branch instead of hardcoding `main`

JAORMX:
This ensures rules are able to leverage the default branch instead of
defaulting to using `main`

### DIFF
--- a/rule-types/github/dockerfile_no_latest_tag.yaml
+++ b/rule-types/github/dockerfile_no_latest_tag.yaml
@@ -29,7 +29,6 @@ def:
   ingest:
     type: git
     git:
-      branch: main
   # Defines the configuration for evaluating data ingested against the given profile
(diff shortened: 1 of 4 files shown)


rule-types/github/dockerfile_no_latest_tag.yaml
```diff
@@ -29,7 +29,6 @@ def:
   ingest:
     type: git
     git:
```
eleftherias, line 31: I'm fairly sure leaving this empty errors when creating the ruletype. It should be an empty object instead
```suggestion
    git: {}
```

JAORMX: It should accept nil, at least that's how I've been creating rule types lately.

eleftherias: You're right

eleftherias (approved):

PR #185, opened 2024-10-16
Introduce tests for rule types

JAORMX:
This uses Minder's testkit in order to create tests for rule types. A
sample is given for the `license` and `secret_scanning_enabled` rule
types.


rule-types/github/secret_scanning.test.yaml
```diff
@@ -0,0 +1,50 @@
...
+        owner: "coolhead"
+        name: "haze-wave"
+    http:
+      body_file: enabled.json
+  - name: "Should have secret scanning enabled for private repo"
+    def:
+      skip_private_repos: false
+    params: {}
+    expect: "pass"
+    entity: *test-repo
```
jhrozek, line 18: Does this refer to the test-repo defined above?

JAORMX: that's right, I'm using a YAML anchor here.

rules_test.go
```diff
@@ -0,0 +1,308 @@
...
+// The Entity Key in EntityVersionWrapper requires a custom unmarshaler
+// Version and Type can be parsed as is.
+// The Entity field depends on the Type field.
+func (e *EntityVersionWrapper) UnmarshalYAML(value *yaml.Node) error {
+	var entity map[string]any
+	if err := value.Decode(&entity); err != nil {
+		return err
+	}
+
+	e.Type = entity["type"].(string)
```
jhrozek, line 59: If type is missing or not a string, then this would panic right?

jhrozek (dismissed):
The tests seem to run and modifying the test data causes failures. I left a nit inline, but since this is test code, feel free to go ahead and merge.

jhrozek (approved):

PR #195, opened 2024-10-22
Add the workflow_no_pull_request_target rule with tests

jhrozek:
Adds a ruletype that checks that a workflow does not use the
`pull_request_target` trigger.

Depends on https://github.com/mindersec/minder/pull/4793

### DIFF
--- a/rule-types/github/workflow_no_pull_request_target.yaml
+++ b/rule-types/github/workflow_no_pull_request_target.yaml
@@ -1,0 +1,63 @@
+---
+version: v1
+release_phase: beta
+type: rule-type
+name: workflow_no_pull_request_target
+display_name: Ensure GitHub Actions workflows do not use the pull_request_target event
+short_failure_message: GitHub Actions workflows use the pull_request_target event
+severity:
+  value: high
+context:
+  provider: github
+description: |
+  Alerts on GitHub Actions workflows that use the pull_request_target event.
+
+  The pull_request_target event allows GitHub Actions workflows to run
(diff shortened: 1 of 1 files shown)


rule-types/github/workflow_no_pull_request_target.yaml
```diff
@@ -0,0 +1,64 @@
...
+  Alerts on GitHub Actions workflows that use the pull_request_target event.
+
+  The pull_request_target event allows GitHub Actions workflows to run
+  on pull requests from forks. This can be a security risk, as the event may,
+  if used improperly, allow untrusted code to run in the repository.
+
+  For more information, see [GitHub's
+  documentation](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_target).
+guidance: |
+  Ensure that GitHub Actions workflows do not use the pull_request_target event.
```
jhrozek, line 22: @eleftherias Would you mind checking if the descriptio, guidance and error reporting match what we expect rules to provide these days towards the UIs?

I admit I wasn't following your work on this lately. Don't feel like you need to review the code as well (unless you want to!)

eleftherias: That all looks good @jhrozek

jhrozek:
Resubmitted without the tests until we have the harness that runs them merged.
(btw I made sure this rule works in Stacklok cloud prod and therefore that the JQ/rego dependencies have been deployed)

rdimitrov (approved):

PR #192, opened 2024-10-22
Introduce an OpenSSF Scorecard profile

ethomson:
One of _my_ goals using Minder is to improve my repository's OpenSSF Scorecard score, since Minder can automatically remediate findings. Minder does not yet have rules to evaluate the full scope of the OpenSSF Scorecard, nor remediate it, but it will still help detect and improve findings.

PR #200, opened 2024-10-24
Add a ruletype that ensures that at least one workflow is triggered by pull_request

jhrozek:
This can be indicative of having a CI/CD pipeline

### DIFF
--- a/rule-types/github/workflow_pull_request.yaml
+++ b/rule-types/github/workflow_pull_request.yaml
@@ -1,0 +1,55 @@
+---
+version: v1
+release_phase: beta
+type: rule-type
+name: workflow_pull_request
+display_name: Ensure there exists a GitHub Actions workflow that uses the pull_request event
+short_failure_message: GitHub Actions workflows do not use the pull_request event
+severity:
+  value: high
+context:
+  provider: github
+description: |
+  Alerts if there are no GitHub Actions workflows that use the pull_request event.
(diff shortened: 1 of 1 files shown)


jhrozek:
@ethomson 👀

ethomson:
:excellent:

eleftherias:
@jhrozek Is there a story this should be tied to? (genuine question, not implying you should create a story)

jhrozek:
> @jhrozek Is there a story this should be tied to? (genuine question, not implying you should create a story)

There's a story of @ethomson pinging me on Slack asking me to do this :-P

No, really, this is related to PR #192 - having an extra ruletype would make the scorecard profile more complete. And since I was already working on PR #195 which is almost the same (just one is positive and one is negative and the event string is different), I was a straightforward thing to do.

eleftherias (approved):

jhrozek:
btw we either need to fix the tests (PR #199 ) or I need to remove the test implementation for now from the PR

jhrozek:
resubmitted without tests so we can merge faster - I'll submit the tests separately once we have the harness to run them merged.

jhrozek:
(btw I made sure this rule works in Stacklok cloud prod and therefore that the JQ/rego dependencies have been deployed)

rdimitrov (approved):

PR #214, opened 2024-11-15
Remove `required` branch from branch protection rules and use default

JAORMX:
This makes it so that branch protection rules no longer have a
`required` parameter. Instead, this is replaced by a default value of
`""`


rdimitrov (approved):

rdimitrov:
@JAORMX - Let's merge it once we confirm the CI job passes successfully, wdyt? 👍

PR #217, opened 2024-12-11
Add data sources and rule types for OSV and Sonatype OSS index

JAORMX:
These two are samples of how to use data sources in your rule types.


rdimitrov (approved):
🚀

PR #222, opened 2024-12-12
Rename osi DS -> spdx-license-list

puerco:
This commit renames the "osi" license list to reference the source of the license data.

/cc @blkt @rdimitrov 

Signed-off-by: Adolfo García Veytia (puerco) <puerco@stacklok.com>

### DIFF
--- a/data-sources/spdx-license-list.yaml
+++ b/data-sources/spdx-license-list.yaml
@@ -1,6 +1,6 @@
 version: v1
 type: data-source
-name: osi
+name: "spdx-license-list"
 context: {}
 rest:
   def:


rdimitrov (approved):

data-sources/spdx-license-list.yaml
```diff
@@ -1,6 +1,6 @@
 version: v1
 type: data-source
-name: osi
```
blkt, line 3: I would name it just `spdx`, as we might add more functions to it.

JAORMX (approved):

data-sources/spdx-license-list.yaml
JAORMX: I agree with this, let's take @blkt 's recommendation

blkt:
We renamed the datasource to `spdx` in another PR.

PR #227, opened 2024-12-17
Add a ruletype for enforcing a file being present in a repo

rdimitrov:
The following PR creates a ruletype that lets you ensure you have a certain file in your repository, but also allows you to specify its expected content.

> This rule is useful for enforcing the presence of various files in the repository, such as LICENSE, README, CONTRIBUTING,
> CODE_OF_CONDUCT, Dependabot configuration files and many more. It can also be used to enforce the presence of specific
> content in the file so you know that the file is not just a placeholder.

Depends on https://github.com/mindersec/minder/pull/5216

This ruletype also implements [OSPS-DO-02](https://baseline.openssf.org/#osps-do-02)


rule-types/common/enforce_file.yaml
```diff
@@ -0,0 +1,89 @@
...
+      content:
+        type: string
+        description: |
+          The content to enforce in the file.
+          For example, the content of the LICENSE file.
+          
+          Leave "" to only check for the presence of the license file.
+    required:
+      - file
+      - content
```
evankanderson, line 39: Question: right now, it looks like you'd end up needing to specify N copies of the rule for N different files.

What would you think of having a schema like the following?

```suggestion
      files:
        type: object
        additionalProperties:
          type: string
        description: |
          A mapping from file path to expected contents in the repository
          
          Set expected contents to "" (empty string) to check for file existence without
          checking file contents.
        default: 
    required:
      - files
```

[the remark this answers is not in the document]
evankanderson: (i.e. can we treat missing as checking the file existence but not contents.)

rule-types/common/enforce_file.yaml
```diff
@@ -0,0 +1,89 @@
...
+        package minder
+
+        import future.keywords.if
+
+        default allow := false
+        fileStr := trim_space(file.read(input.profile.file))
+
+        allow if {
+            # Read the file and check if it contains the content
+            # contains(fileStr, input.profile.content)
```
evankanderson, line 59: This comment (and the one on line 70) seem confusing.

rule-types/common/enforce_file.yaml
```diff
@@ -0,0 +1,89 @@
...
+      title: "Ensure {{.Profile.file }} exists with the expected content"
+      body: |
+        This is a Minder automated pull request.
+        
+        This pull request ensures that this repository contains the file {{.Profile.file}} with the expected content set by your organization.
+      contents:
+        - path: "{{.Profile.file}}"
+          action: replace
+          content: |
+            {{.Profile.content}}
```
evankanderson, line 85: Ah!  This is a (current) reason not to use the `object` formulation -- we don't have a way to iterate over all the keys in the `Profile.files` map, correct?

evankanderson (commented):
A few nits, but seems good overall.

rule-types/common/enforce_file.yaml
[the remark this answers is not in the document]
rdimitrov: I was unsure about which of the 2 is more explicit from the user's point of view. Stating it now I think I agree that making `content` optional makes more sense 👍

rule-types/common/enforce_file.yaml
rdimitrov: Oh, I must have missed to delete that, they are leftover parts from writing/debugging the rule. Thanks 👍

rule-types/common/enforce_file.yaml
rdimitrov: Exactly 👍 Of course, we can try to do that if we want. I was also thinking if I should try and make the path field optional for the pull request remediation type, but at the same time I like explicitly scoping it so you don't accidentally push another change. My concern for not looking at these is both will be considered breaking changes so I think it'll be out of scope for this one.

rule-types/common/enforce_file.yaml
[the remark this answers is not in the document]
rdimitrov: Fixed

rule-types/common/enforce_file.yaml
```diff
@@ -0,0 +1,101 @@
...
+    rego:
+      type: deny-by-default
+      def: |
+        package minder
+
+        import future.keywords.if
+
+        default allow := false
+        default skip := false
+        fileStr := trim_space(file.read(input.profile.file))
```
blkt, line 60: **concern:** I'm afraid this can't be used to implement `OSPS-DO-02` since it requires checking the existence of `CONTRIBUTING/` folder as well.

How complex would it be to support folders as well?

rdimitrov: Let me see what I can do 👍

blkt: I guess the main difficulty is that `input.profile.file` is also used for the remediation, and in the case of a folder it's harder to figure out what's the right thing to do (i.e. one can't always default to creating a file).

On the other hand, OSSF is not currently interested in remediation.

blkt (approved):

PR #228, opened 2024-12-17
Add ruletype for Renovate GitHub Action

eleftherias:


rdimitrov:
What do you think about making this more generic? 

Instead of having several ruletypes for each action we want to check against, I'd propose to have a generic one extended by a parameter through which you can specify the action. 

Perhaps we can even repurpose the license ruletype into this new ruletype. Currently it allows you to specify a file path and check if a given content is there. Extending it to support a directory of files should be enough to cover the above use case and much more.

JAORMX:
wow! I had no idea that Renovate had a github action. Normally, folks install the Renovate App, and then install a configuration file in the repo: https://docs.renovatebot.com/config-overview/ https://docs.renovatebot.com/configuration-options/

Perhaps we can have yet another rule for this :smile:

mesembria:
> What do you think about making this more generic?
> 
> Instead of having several ruletypes for each action we want to check against, I'd propose to have a generic one extended by a parameter through which you can specify the action.
> 
> Perhaps we can even repurpose the license ruletype into this new ruletype. Currently it allows you to specify a file path and check if a given content is there. Extending it to support a directory of files should be enough to cover the above use case and much more.

I think that's a good idea for a quick, first pass of a rule to check for GitHub actions. But if you wanted to enhance the rule, or make it more specific, the generic approach might have limitations. For example, with this Renovate rule we also discussed checking for the required config file and checking the schedule as a next step.

rdimitrov:
My concern for raising this is this may lead us to the same situation we are with the existing Dependabot ruletype. At the same time I see that having more examples of rules like that could help us to understand how a generic one might look like and would it be useful.

rule-types/github/renovate_github_action.yaml
```diff
@@ -0,0 +1,50 @@
...
+        default message := "Renovate GitHub action is not configured"
+        default allow := false
+        allow if {
+          # check that there is a renovate action
+          "renovatebot/github-action" in actions
+        }
+  # Defines the configuration for alerting on the rule
+  alert:
+    type: security_advisory
+    security_advisory: {}
```
JAORMX, line 50: nit: Should we remove the `security_advisory` alerts?

JAORMX (commented):
LGTM

PR #230, opened 2024-12-18
Add Yelp Secret Scanner Action validation rule

teodor-yanev:


JAORMX (commented):
Looks good! Can we have some tests for this?

JAORMX (approved):

PR #232, opened 2024-12-19
Add check for OSPS-DO-03: repo contains end-user documentation

evankanderson:
This is a proposed implementation of https://baseline.openssf.org/#osps-do-03

Ideally, remediation would include creating or updating SECURITY-INSIGHTS.md, but we don't quite have those capabilities yet.  I've included a sample of what we _are_ able to do at the moment, but we'd want to be able to feed the set of discovered documentation locations into the documentation field in security-insights if not set.


rule-types/github/osps-do-03.yaml
```diff
@@ -0,0 +1,97 @@
...
+  This rule attempts to locate user guides from several project documentation sources.
+
+  Currently, this rule checks the following:
+
+  * The GitHub repository's public link
+  * A `docs` directory in the default branch of the repository with .md, .rst, .html or .txt files
+  * A `README.md` file containing preformatted text (triple-backtick) or the headings
+   "usage" or "getting started"
+
+  For more information, see [OpenSSF Security Baseline](https://baseline.openssf.org/#osps-do-03).
```
JAORMX, line 24: The guidance is not meant to describe the rule type itself, but instead, to indicate what to do in case of failure. This should be moved to the description and we should tell the user what to do in case this fails.

evankanderson: Does that apply to the last sentence, or the whole current guidance?

JAORMX: This while guidance felt more like a description

evankanderson: Does this read better?

rule-types/github/osps-do-03.yaml
```diff
@@ -0,0 +1,103 @@
+---
+version: v1
+release_phase: alpha
+type: rule-type
+name: osps-do-03
```
blkt, line 5: **nit:** we usually name rules with underscores, but, honestly, I like it more this way as well.

rule-types/github/osps-do-03.yaml
blkt, line 1: **note:** we decided to move rule types related to OSPS under `security-baseline/rule-type`, I can move this after merge as part of the work I'm doing.

blkt (dismissed):

rule-types/github/osps-do-03.yaml
```diff
@@ -0,0 +1,103 @@
...
+        # TODO: these should trigger a remediation to put them in
+        # SECURITY-INSIGHTS.yaml, but also still pass the check(?)
+
+        allow if {
+          # Check the GitHub homepage link
+          out = minder.datasource.ghapi.repo({
+            "owner": input.properties["github/repo_owner"],
+            "repo": input.properties["github/repo_name"]
+          })
+          out.homepage != ""
```
blkt, line 67: **issue:** this needs an additional indirection IIRC.
```suggestion
          out.body.homepage != ""
```

rule-types/github/osps-do-03.yaml
evankanderson: We moved `security-baseline` to be the other way.  \o/

rule-types/github/osps-do-03.yaml
evankanderson: I've moved this and it should be ready for review.

security-baseline/rule-types/github/osps-do-03.yaml
```diff
@@ -0,0 +1,103 @@
...
+      def: |
+        package minder
+
+        import rego.v1
+
+        default allow := false
+
+        repo := sprintf("%s/%s", [])
+
+        allow if {
```
eleftherias, line 51: We should mention the SECURITY-INSIGHTS.yaml file in the rule guidance and description.

security-baseline/rule-types/github/osps-do-03.yaml
```diff
@@ -0,0 +1,103 @@
...
+
+        readme := file.read("README.md")
+        allow if {
+          # Check the README.md file for preformatted text after the first line
+          regex.match("\n *```", readme)
+        }
+        allow if {
+          regex.match("\n#+ (?i:Usage|Getting Started)", readme)
+        }
+  remediate:
```
eleftherias, line 87: Is there value in keeping the remediation? It's not very flexible at the moment and may link to the wrong file.

security-baseline/rule-types/github/osps-do-03.yaml
evankanderson: Done.

security-baseline/profiles/security-baseline-level-1.yaml
```diff
@@ -32,7 +32,13 @@ repository:
   # OSPS-BR-09: GitHub hosted releases have this, check for links in release text?
 
   # OSPS-DO-03: Ensure user guides for all basic functionality
+  - name: osps-do-03
+    type: osps-do-03
+    def: {}
   # OSPS-DO-05: Project documentation has a mechanism for reporting defects
+  - name: osps-do-05
```
eleftherias, line 39: The ruletype for osps-do-05 doesn't exist yet, were you going to add it as part of this PR?

security-baseline/rule-types/github/osps-do-03.yaml
evankanderson: Good point.  I think it has some interesting insights on features to add for remediation, but I've commented it out for now.

security-baseline/profiles/security-baseline-level-1.yaml
evankanderson: Oops, I started to, then decided to separate things, but forgot to save the file.

eleftherias (approved):

PR #233, opened 2024-12-19
Talisman pre-commit hook secrets scanning

peppescg:
co-author @kantord

### DIFF
--- a/rule-types/github/talisman_secrets_scanning.yaml
+++ b/rule-types/github/talisman_secrets_scanning.yaml
@@ -40,10 +40,13 @@ def:
         
         parsed_data := parse_yaml(precommit)
 
-        allow if {
-          some repo_id
+         allow if {
+          some repo_id, hook_id
           repo_data := parsed_data.repos[repo_id]
           endswith(repo_data["repo"], "https://github.com/thoughtworks/talisman")
+          talisman_hooks = repo_data["hooks"]
+          talisman_hooks[hook_id].id == "talisman-commit"
+          talisman_hooks[hook_id].entry == "cmd --githook pre-commit"
         }
 
         message := "" if allow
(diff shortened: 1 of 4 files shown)


rule-types/github/talisman_secrets_scanning.yaml
```diff
@@ -0,0 +1,52 @@
...
+        
+        # pre-commit hook
+        precommit := file.read(".pre-commit-config.yaml")
+        
+        parsed_data := parse_yaml(precommit)
+
+        allow if {
+          some repo_id
+          repo_data := parsed_data.repos[repo_id]
+          endswith(repo_data["repo"], "https://github.com/thoughtworks/talisman")
```
JAORMX, line 46: I see we're checking for the hook coming from this repo to be set. Do we want to check the hook entries as well though?

I mean:
```
      - id: talisman-commit
        entry: cmd --githook pre-commit
```

We can iterate the hooks and verify those.

kantord: yes, this is a good idea...

I was wondering though. If we add such complex business logic for each rule concerning pre-commit, and then copy paste the exact same code for every imaginable pre-commit hook, then we are creating a maintenance nightmare :thinking: 

so I guess we could add a helper function in minder for this (like we have for github actions), or else we could convert this rule into a parametric rule where you have to specify which hook you want to enforce?

kantord: > convert this rule into a parametric rule where you have to specify which hook you want to enforce?

which kinda raises the question if we can somehow support "partial application" of rule types, i. e. we make a generic rule type that takes 1 argument as required configuration, and we build other rule types that re-use that parametric rule type, but don't require any parameters :thinking:

JAORMX: this is a really good idea. I haven't really figured out how this would look like in terms of implementation, but I can definitely see us supporting libraries... Let me give this some thought. Can you add an entry in the Gaps tracker?

kantord: > Gaps tracker?

what is that

JAORMX (approved):

PR #238, opened 2024-12-19
Introduce OpenSSF Best Practices Badge rule

ethomson:
Add a rule that ensures that the project has an OpenSSF Best Practices badge at the specified level (defaulting to: "passing"). This is driven by the OpenSSF Best Practices data source, which queries bestpractices.dev.

Add the new rule type to the OpenSSF Scorecard profile.

### DIFF
--- a/rule-types/github/openssf_bestpractices_badge.yaml
+++ b/rule-types/github/openssf_bestpractices_badge.yaml
@@ -1,0 +1,77 @@
+---
+version: v1
+release_phase: alpha
+type: rule-type
+name: openssf_bestpractices
+display_name: Verifies the project has earned an OpenSSF Best Practices Badge
+short_failure_message: OpenSSF Best Practices Badge is missing
+severity:
+  value: low
+context:
+  provider: github
+description: |
+  Verifies that a repository contains an OpenSSF Best Practices badge at
+  the specified level.
+
+  This rule type checks for the existence of a image declared in Markdown,
+  referencing the project's badge on the Best Practices badge site.
+guidance: |
+  The [OpenSSF Best Practices Badge](https://www.bestpractices.dev/en)
+  program allows open source projects to show that they follow best
+  security practices.
+def:
+  in_entity: repository
+  # Defines the schema for writing a rule with this rule being checked
+  # In this case there are no settings that need to be configured
+  rule_schema:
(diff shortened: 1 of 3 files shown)


rule-types/github/openssf_bestpractices_badge.yaml
```diff
@@ -0,0 +1,77 @@
...
+          project_id := badge[0][1]
+
+          badge_data := minder.datasource.openssf_bestpractices.lookup({"id": project_id })
+
+          levels[badge_data.body.badge_level] >= levels[input.profile.level]
+        }
+  # Defines the configuration for alerting on the rule
+  alert:
+    type: security_advisory
+    security_advisory: {}
```
JAORMX, line 77: nit: Should we stop adding the `security_advisory` alert? we probably want to deprecate this anyway.

rule-types/github/openssf_bestpractices_badge.yaml
```diff
@@ -0,0 +1,77 @@
+---
+version: v1
+release_phase: alpha
+type: rule-type
+name: openssf_bestpractices
```
JAORMX, line 5: This must match the file name.

rule-types/github/openssf_bestpractices_badge.yaml
```diff
@@ -0,0 +1,77 @@
...
+description: |
+  Verifies that a repository contains an OpenSSF Best Practices badge at
+  the specified level.
+
+  This rule type checks for the existence of a image declared in Markdown,
+  referencing the project's badge on the Best Practices badge site.
+guidance: |
+  The [OpenSSF Best Practices Badge](https://www.bestpractices.dev/en)
+  program allows open source projects to show that they follow best
+  security practices.
```
JAORMX, line 21: This guidance looks a little off. The intent is to tell people what to do if they fail this check.

rule-types/github/openssf_bestpractices_badge.yaml
ethomson: Yes _but_ I think that it makes sense to burn it down so that the rules are consistent with each other, instead of having a few that are one way and a few that are another.

ethomson:
Updated in #243

PR #241, opened 2024-12-19
Semgrep pre-commit hook

peppescg:


rule-types/github/semgrep_secrets_scanning.yaml
```diff
@@ -0,0 +1,53 @@
...
+release_phase: alpha
+type: rule-type
+name: semgrep_secrets_scanning
+display_name: Enable Semgrep Pre-commit hooks for detecting secrets
+short_failure_message: Semgrep Pre-commit hook is not configured for the repository
+severity:
+  value: medium
+context: {}
+description: |
+  Verifies that Semgrep Pre-commit hook is configured via a GitHub action for the repository, and remediate
```
kantord, line 12:
```suggestion
  Verifies that Semgrep Pre-commit hook is configured for the repository, and remediate
```

rule-types/github/semgrep_secrets_scanning.yaml
```diff
@@ -0,0 +1,53 @@
...
+    type: git
+    git: {}
+  eval:
+    type: rego
+    rego:
+      type: deny-by-default
+      def: |
+        package minder
+        import future.keywords.if
+        import future.keywords.every
```
evankanderson, line 31: I think `import rego.v1` brings you both of these.

```suggestion
        import rego.v1
```

rule-types/github/semgrep_secrets_scanning.yaml
```diff
@@ -0,0 +1,53 @@
...
+
+        allow if {
+          some repo_id, hook_id
+          repo_data := parsed_data.repos[repo_id]
+          endswith(repo_data["repo"], "https://github.com/semgrep/pre-commit")
+          semgrep_hooks = repo_data["hooks"]
+          semgrep_hooks[hook_id].id == "semgrep"
+        }
+
+        message := "" if allow
```
evankanderson, line 50: I believe message is only used when `allow` is false, so you can just do:

```suggestion
        message := "Semgrep pre-commit hook is not configured for the repository"
```

rule-types/github/semgrep_secrets_scanning.yaml
```diff
@@ -0,0 +1,53 @@
...
+        precommit := file.read(".pre-commit-config.yaml")
+        
+        parsed_data := parse_yaml(precommit)
+
+        allow if {
+          some repo_id, hook_id
+          repo_data := parsed_data.repos[repo_id]
+          endswith(repo_data["repo"], "https://github.com/semgrep/pre-commit")
+          semgrep_hooks = repo_data["hooks"]
+          semgrep_hooks[hook_id].id == "semgrep"
```
evankanderson, line 47: It looks like Semgrep has 4 configurations: `semgrep`, `semgrep-ci`, `semgrep-docker`, and `semgrep-docker-develop`.  It looks like this requires the `semgrep` one and not (for example) the CI one.  Is this usage intentional?

evankanderson (approved):
A few small nits, but approving in case they aren't needed.

PR #246, opened 2024-12-19
add rule to require pre-commit to be added to github actions

kantord:


rule-types/github/pre_commit_github_action.test.yaml
```diff
@@ -0,0 +1,13 @@
...
+    params: {}
+    expect: "pass"
+    git:
+      repo_base: good
+  - name: "Should not have pre-commit enabled"
+    def: {}
+    params: {}
+    expect: "fail"
+    git:
+      repo_base: bad
```
JAORMX, line 13: you're missing adding these directories

kantord: yup, I am working on it

rule-types/github/pre_commit_github_action.yaml
```diff
@@ -0,0 +1,41 @@
...
+name: pre_commit_github_action
+display_name: Require running `pre-commit` using GitHub Actions
+short_failure_message: "`https://github.com/pre-commit/action` action is not configured"
+severity:
+  value: medium
+context: {}
+description: |
+  Verifies that `pre-commit` is configured via a GitHub action for the repository.
+guidance: |
+  Run your `pre-commit` checks in GitHub Actions as well.
```
JAORMX, line 14: Can we point folks to documentation on how to do this?

PR #249, opened 2024-12-19
Added rule to check for Trufflehog Github action

mesembria:
Added a rule to check for TruffleHog Github action.


JAORMX:
@mesembria oops! Seems you accidentally checked in some extra rule types for repo visibility

mesembria:
Not sure how I managed that, but it should be cleaned up now. Thanks for the catch @JAORMX!

PR #250, opened 2024-12-19
Gitleaks GitHub Action Rule

dussab:
Checks to make sure Gitleaks GitHub action is enabled to prevent the leakage of passwords or keys.

### DIFF
--- a/rule-types/github/gitleaks_github_action.yaml
+++ b/rule-types/github/gitleaks_github_action.yaml
@@ -3,15 +3,15 @@ version: v1
 release_phase: alpha
 type: rule-type
 name: gitleaks_github_action
-display_name: Enable Gitleak to prevent leakage of passwords or keys
-short_failure_message: Gitleak is not configured via a GitHub action
+display_name: Enable Gitleaks to prevent leakage of passwords or keys
+short_failure_message: Gitleaks is not configured via a GitHub action
 severity:
   value: medium
 context: {}
 description: |
-  Verifies that Gitleak is configured via a GitHub action for the repository.
+  Verifies that Gitleaks is configured via a GitHub action for the repository.
 guidance: |
-  Ensure that Gitleak is configured and enabled for the repository.
+  Ensure that Gitleaks is configured and enabled for the repository.
   Gitleaks is an open-source secret scanner for git repositories, files, and directories.
   For more information, see the [GitHub Action Gitleaks](https://github.com/gitleaks/gitleaks-action) documentation.
 def:


rule-types/github/gitleaks_github_action.yaml
```diff
@@ -0,0 +1,43 @@
...
+short_failure_message: Gitleak is not configured via a GitHub action
+severity:
+  value: medium
+context: {}
+description: |
+  Verifies that Gitleak is configured via a GitHub action for the repository.
+guidance: |
+  Ensure that Gitleak is configured and enabled for the repository.
+  Gitleaks is an open-source secret scanner for git repositories, files, and directories.
+  For more information, see the [GitHub Action Gitleaks](https://github.com/gitleaks/gitleaks) documentation.
```
JAORMX, line 16: nit: should we show a sample of how a workflow with this action looks like?

rule-types/github/gitleaks_github_action.yaml
```diff
@@ -0,0 +1,43 @@
...
+      def: |
+        package minder
+        
+        import rego.v1
+
+        actions := github_workflow.ls_actions("./.github/workflows")
+
+        default message := "Gitleaks GitHub action is not configured"
+        default allow := false
+        allow if {"gitleaks/gitleaks-action" in actions
```
JAORMX, line 38: Try https://docs.styra.com/opa/rego-by-example/keywords/contains

rule-types/github/gitleaks_github_action.yaml
dussab: I updated the guidance and linked directly to the github action repo. Do you think i should add the Github action configuration directly to the guidance section?

rule-types/github/gitleaks_github_action.yaml
dussab: Updated to use contains keyword.

rule-types/github/gitleaks_github_action.yaml
```diff
@@ -0,0 +1,43 @@
+---
+version: v1
+release_phase: alpha
+type: rule-type
+name: gitleaks_github_action
+display_name: Enable Gitleak to prevent leakage of passwords or keys
```
evankanderson, line 6: s/Gitleak/Gitleaks/

https://github.com/gitleaks/gitleaks seems to be consistently plural

rule-types/github/gitleaks_github_action.yaml
```diff
@@ -0,0 +1,43 @@
...
+      def: |
+        package minder
+        
+        import rego.v1
+
+        actions := github_workflow.ls_actions("./.github/workflows")
+
+        default message := "Gitleaks GitHub action is not configured"
+        default allow := false
+        allow if contains(actions[_], "gitleaks/gitleaks-action")
```
evankanderson, line 38: Do you want to check that the workflow is `on` for both `pull_request` and `push` events?

evankanderson: Also, do you want to add a remediation using the [usage example](https://github.com/gitleaks/gitleaks-action?tab=readme-ov-file#usage-example) from either the action or the repo?

rule-types/github/gitleaks_github_action.yaml
dussab: Good eyes.  Updated text to be consistently plural

evankanderson (approved):

evankanderson:
Actually, tests are failing, so we'll need to fix that first.

PR #254, opened 2024-12-20
Add a ruletype that checks the issue tracker status

rdimitrov:
The following ruletype checks the issue tracker status (whether it is enabled or disabled).

It should satisfy [OSPOS-DO-05](https://baseline.openssf.org/#osps-do-05)


rule-types/github/repo_issues.yaml
```diff
@@ -0,0 +1,66 @@
...
+def:
+  in_entity: repository
+  rule_schema:
+    type: object
+    properties:
+      has_issues:
+        type: boolean
+        description: "Whether the repository should have issues enabled (true) or disabled (false)."
+    required:
+      - has_issues
```
JAORMX, line 44: I wonder if we want to have an explicit rule that verifies that the issue tracker is indeed enabled. This would make the rule simpler since it would not have a toggle, and we could tailor the guidance to that case. wdyt?

rdimitrov: Sounds good 👍 So splitting this to 2 separate rules, i.e. `issues_enabled` and `issues_disabled`?

rdimitrov: on it 🚀

rdimitrov:
Closing this in favour of https://github.com/mindersec/minder-rules-and-profiles/pull/255 and #256

PR #255, opened 2024-12-20
Add a ruletype that checks if issues are enabled

rdimitrov:
The following ruletype checks that the issue tracker status is enabled and remediates it if needed.

It should satisfy [OSPOS-DO-05](https://baseline.openssf.org/#osps-do-05)


JAORMX (approved):

rule-types/github/repo_issues_enabled.yaml
```diff
@@ -0,0 +1,49 @@
...
+  remediate:
+    type: rest
+    rest:
+      method: PATCH
+      endpoint: "/repos/{{.Entity.Owner}}/{{.Entity.Name}}"
+      body: |
+        { "has_issues": true }
+  alert:
+    type: security_advisory
+    security_advisory: {}
```
evankanderson, line 49: I suspect we don't want a security alert for this?

PR #256, opened 2024-12-20
Add a ruletype that checks if issues are disabled

rdimitrov:
The following ruletype checks that the issue tracker status is disabled and remediates it if needed.

It should relate to [OSPOS-DO-05](https://baseline.openssf.org/#osps-do-05)


JAORMX (approved):

evankanderson:
> The following ruletype checks that the issue tracker status is disabled and remediates it if needed.
> 
> It should relate to [OSPOS-DO-05](https://baseline.openssf.org/#osps-do-05)

I think OSPS-DO-05 is actually encouraging having issues _enabled_, or clearly linking to an external tracker.

rdimitrov:
> > The following ruletype checks that the issue tracker status is disabled and remediates it if needed.
> > It should relate to [OSPOS-DO-05](https://baseline.openssf.org/#osps-do-05)
> 
> I think OSPS-DO-05 is actually encouraging having issues _enabled_, or clearly linking to an external tracker.

You're right 👍 I added a comment that the `enabled` one should satisfy it. My thought how this one might be related was the case where there's an external issue tracker and the one from GitHub should be explicitly disabled. In any case it's not part of the spec so I agree it's perhaps more misleading than helpful 👍

PR #259, opened 2025-01-08
Fix indirect vulnerabilities by bumping minder and protobuf

rdimitrov:
The following PR bumps direct dependencies which should resolve 1 dependabot-raised vulnerability.


eleftherias:
Weirdly the workflow looks like it's passing, but the test job is actually fails with `./rules_test.go:150:51: not enough arguments in call to rtengine.NewRuleTypeEngine`

rdimitrov:
> Weirdly the workflow looks like it's passing, but the test job is actually fails with `./rules_test.go:150:51: not enough arguments in call to rtengine.NewRuleTypeEngine`

Thanks for noticing this, fixed it! 👍

eleftherias (approved):

PR #261, opened 2025-01-08
Add `permissive_license` rule type.

blkt:
This rule type checks that the license detected by GitHub is approved by either OSI or FSF. It uses two data sources, one to call GitHub API to get the SPDX identifier of the license, and another one to get the updated list of licenses approved by from SPDX repository.

This rule can be used to implement `OSPS-LE-02`.

P.S.: the name is ugly, please advise. 😅

### DIFF
--- a/rule-types/github/permissive_license.yaml
+++ b/rule-types/github/permissive_license.yaml
@@ -19,11 +19,22 @@ def:
   in_entity: repository
   rule_schema: {}
   ingest:
-    type: git
+    type: rest
+    rest:
+      # This is the path to the data source. Given that this will evaluate
+      # for each repository in the organization, we use a template that
+      # will be evaluated for each repository. The structure to use is the
+      # protobuf structure for the entity that is being evaluated.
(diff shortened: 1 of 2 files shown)


rule-types/github/permissive_license.yaml
```diff
@@ -0,0 +1,53 @@
...
+  recognized and legally enforceable open source software license.
+guidance: |
+  Ensure that the project’s source code is distributed under a
+  recognized and legally enforceable open source software license,
+  providing clarity on how the code can be used and shared by others.
+def:
+  in_entity: repository
+  rule_schema: {}
+  ingest:
+    type: git
```
JAORMX, line 22: Is this needed? Why not use the REST ingest directly?

rule-types/github/permissive_license.yaml
```diff
@@ -0,0 +1,53 @@
...
+  eval:
+    type: rego
+    data_sources:
+      - name: ghapi
```
JAORMX, line 26: The usage of this data source seems unnecessary, you could simply use the REST ingest.

JAORMX (approved):

PR #267, opened 2025-01-13
Add rule type to verify that a lockfile exists in a Python project

JAORMX:
The idea is to have a catch-all rule that helps organizations enforce
the usage of lockfiles.


rule-types/common/python_lock_file_exists.yaml
```diff
@@ -0,0 +1,98 @@
+---
+version: v1
+release_phase: alpha
+type: rule-type
+name: python_lock_file_exists
+display_name:  Python lock file exists
+short_failure_message:  A Python lock file is missing
```
rdimitrov, line 7: nit: Would it sound better if this is named like `use pinned dependencies in python` or something else that is more descriptive?

rdimitrov: Actually I meant this for all three not just the failure message

rdimitrov (approved):

PR #275, opened 2025-01-17
Add osps-qa-02 rule

teodor-yanev:
Fixes https://github.com/stacklok/minder-stories/issues/200

We haven't received a confirmation on the OpenSSF channel yet so I made the assumptions based on the official rule definition here https://baseline.openssf.org/#osps-qa-02

They are described in the rule definition's `description` and `guidance` parts.


security-baseline/rule-types/github/osps-qa-02.yaml
```diff
@@ -0,0 +1,59 @@
...
+     [Repository Settings](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/managing-repository-settings/setting-repository-visibility) page.
+  2. Ensure force pushes are disabled in branch protection rules via the
+     [Branch protection settings](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/managing-a-branch-protection-rule).
+  3. Ensure squashing commits is disabled in the repository settings via the
+     [Configure commit squashing](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/configuring-commit-squashing-for-pull-requests).
+def:
+  in_entity: repository
+  rule_schema: {}
+  ingest:
+    type: git
```
JAORMX, line 27: why are you using the `git` ingest if you're not accessing the git repo itself?

data-sources/ghapi.yaml
```diff
@@ -24,3 +24,15 @@ rest:
             type: string
           repo:
             type: string
+    branch_protection:
```
JAORMX, line 27: is this needed? You could just use the REST ingest instead.

security-baseline/rule-types/github/osps-qa-02.yaml
```diff
@@ -0,0 +1,59 @@
...
+          "branch": input.properties["github/default_branch"]
+        })
+
+        allow if {
+          not input.properties["is_private"]
+
+          not branch_protection.body.allow_force_pushes.enabled
+          
+          not repo.body.allow_squash_merge
+        }
```
JAORMX, line 59: You could just leverage the result of the REST ingest for the branch you're targeting and do this.

JAORMX (changes requested):
We arleady have a branch protection rule to check that force pushing is not allowed https://github.com/mindersec/minder-rules-and-profiles/blob/main/rule-types/github/branch_protection_allow_force_pushes.yaml

Why are we introducing a new one and why are we doing it differently?

evankanderson (commented):
Oops, I forgot to submit my changes, which generally agree with Ozz's comments.

teodor-yanev:
> We arleady have a branch protection rule to check that force pushing is not allowed https://github.com/mindersec/minder-rules-and-profiles/blob/main/rule-types/github/branch_protection_allow_force_pushes.yaml
> 
> Why are we introducing a new one and why are we doing it differently?

Some rules in the directory were copy-pasted from rules like e.g. branch\_protection\_allow\_deletions in order to
 (a) be able to change them independently and 
(b) change the name to something descriptive in the scope of Security Baseline.

We generally do not foster this, but in this case we deemed simplicity was preferable to avoiding duplication.

teodor-yanev:
note: squashing was removed as per https://openssf.slack.com/archives/C07DC6TT2QY/p1737165055721979?thread_ts=1737159831.490839&cid=C07DC6TT2QY

data-sources/ghapi.yaml
teodor-yanev: + I also forget that we do not have authentication for data sources and this call would've failed anyway 🙃

JAORMX (approved):

PR #278, opened 2025-01-24
Add rule for OSPS-BR-01

evankanderson:
This implements OSPS-BR-01 as interpreted as "[the scorecards dangerous workflows check](https://github.com/ossf/scorecard/blob/main/docs/checks.md#dangerous-workflow)", which was the previous definition.

We may want to discuss splitting this into multiple checks for different criteria, but that seems separate from the item today.

### DIFF
--- a/rules_test.go
+++ b/rules_test.go
@@ -4,17 +4,18 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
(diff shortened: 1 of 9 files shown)


eleftherias (approved):

evankanderson:
Merged #279, closing this.

PR #279, opened 2025-01-24
Rename DO- rules, inventory remaining rules missing from lv1 profile

evankanderson:
The DO-01 and DO-02 rules were moved into the GV category.  Add additional comments in the profile YAML about when / how the remaining rules will be implemented.


evankanderson:
This includes the changes from #278; review that one first.

eleftherias (approved):

PR #296, opened 2025-01-28
Add ruletype OSPS-LE- 04

teodor-yanev:
Closes https://github.com/stacklok/minder-stories/issues/199

This PR should be merged only after the external license processor is up and running. The URL in the POST used here should then be updated.
For visibility, this is the lib that will do the processing: https://github.com/CycloneDX/license-scanner

### DIFF
--- a/security-baseline/rule-types/github/osps-le-04.yaml
+++ b/security-baseline/rule-types/github/osps-le-04.yaml
@@ -12,7 +12,7 @@ description: |
   Verifies that released software assets include a license file that meets
   the OSI Open Source Definition or the FSF Free Software Definition.
 guidance: |
-  Ensure that your releases include a LICENSE or COPYING file that contains
+  Ensure that your releases include a LICENSE file that contains
   an OSI-approved or FSF-approved license. Common examples include:
   - Apache License 2.0
   - BSD License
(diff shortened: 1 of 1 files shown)


security-baseline/rule-types/github/osps-le-04.yaml
```diff
@@ -0,0 +1,97 @@
...
+          # This should be the servcice processing on the other side:
+          # https://github.com/CycloneDX/license-scanner
+          ident_resp := http.send({
+            "method": "POST",
+            "url": "https://license-identifier-service.example.com/identify",
+            "headers": {
+              "Accept": "application/json",
+              "Content-Type": "application/octet-stream"
+            },
+            "raw_body": license_resp
```
evankanderson, line 81: Doesn't this need to be the `.body` or some other attribute of the http response?

security-baseline/rule-types/github/osps-le-04.yaml
```diff
@@ -0,0 +1,97 @@
...
+          })
+          
+          license_id := ident_resp.license.id
```
evankanderson, line 84: What format do you want this to be.  I'm guessing it will be something like a JSON list of SPDX license strings (to support multiple licenses in a file):

```json
[
  "MIT",
  "Apache-2.0"
]
```

evankanderson: I did this for now; it's in https://github.com/evankanderson/spdx-detector for the moment, but I may move it to StacklokLabs (shouldn't change anything else if I do that).

security-baseline/rule-types/github/osps-le-04.yaml
evankanderson: I meant that `license_resp` is an object, not a string.  I think you want:

```suggestion
            "raw_body": license_resp.raw_body
```

security-baseline/rule-types/github/osps-le-04.yaml
```diff
@@ -0,0 +1,91 @@
...
+          
+          license_id := ident_resp.body[0]
+          
+          resp := minder.datasource.spdx.licenses({})
+          licenses := resp.body.licenses
+          osi := { l.licenseId | l := licenses[_]; l.isOsiApproved }
+          fsf := { l.licenseId | l := licenses[_]; l.isFsfLibre }
+          approved_licenses := osi | fsf
+          
+          count(approved_licenses) != 0
```
evankanderson, line 86: This is just a general assertion, not specifically needed for this rule, right?

security-baseline/rule-types/github/osps-le-04.yaml
```diff
@@ -0,0 +1,90 @@
...
+          license_resp := http.send({
+            "method": "GET",
+            "url": license_file.browser_download_url,
+            "headers": {
+              "Accept": "application/octet-stream"
+            }
+          })
+          
+          # We have this servcice processing on the other side:
+          # https://github.com/CycloneDX/license-scanner
```
evankanderson, line 67:
```suggestion
          # We have a service for this detection:
          # https://github.com/evankanderson/spdx-detector
```

evankanderson (approved):
I'm assuming you've tested this in staging?

security-baseline/rule-types/github/osps-le-04.yaml
```diff
@@ -0,0 +1,90 @@
...
+        }
+
+        violations[{"msg": msg}] {
+          license_file := asset {
+            some asset in input.ingested
+            is_license_file(asset.name)
+          }
+
+          # Exit without violation if no license file found
+          license_file != {}
```
evankanderson, line 56: I don't think you need this; since `license_file` is some entry in `input.ingested` which has a `name` property which passes `is_license_file`, you _know_ it's not an empty object (it must contain a `name` property).

security-baseline/rule-types/github/osps-le-04.yaml
```diff
@@ -0,0 +1,90 @@
...
+        license_pattern := "^license(\\..*)?$"
+        is_license_file(name) if {
+          regex.match(license_pattern, lower(name))
+        }
+
+        violations[{"msg": msg}] {
+          license_file := asset {
+            some asset in input.ingested
+            is_license_file(asset.name)
+          }
```
evankanderson, line 53: This seems simpler as:

```suggestion
        violations[{"msg": msg}] {
          license_file := asset {
            some asset in input.ingested
            startswith(lower(name), "license")
          }
```

evankanderson (dismissed):

security-baseline/rule-types/github/osps-le-04.yaml
```diff
@@ -0,0 +1,90 @@
...
+
+        get_license_info(asset) := {"license_resp": license_resp, "ident_resp": ident_resp} if {
+            license_resp := http.send({
+                "method": "GET",
+                "url": asset.browser_download_url,
+                "headers": {
+                    "Accept": "application/octet-stream"
+                },
+                "enable_redirect": true,
+                "raise_error": true
```
evankanderson, line 52: This is fine, but it's also [the default for `http.send`](https://www.openpolicyagent.org/docs/latest/policy-reference/#http).

evankanderson (approved):

PR #307, opened 2025-01-31
Add ruletype for OSPS-QA-03

### LINKED ISSUE #289: Implement Security Baseline criterion `OSPS-QA-03`
**Criterion**: All [released software assets](https://baseline.openssf.org/#released-software-asset) MUST be delivered with a machine-readable list of all direct and transitive internal software dependencies with their associated version identifiers.

**Maturity Level**: 2

**Rationale**: Provide transparency and accountability for the project’s dependencies, enabling users and [contributors](https://baseline.openssf.org/#contributor) to understand the software’s dependencies and versions.

**Details**: This may take the form of a software bill of materials (SBOM) or a dependency file that lists all direct and transitive dependencies such as package.json, Gemfile.lock, or go.sum.

It is recommended to use a CycloneDX or SPDX file that is auto-generated at build time by a tool that has been vetted for accuracy. This enables users to ingest this data in a standardized approach alongside other projects in their environment.

eleftherias:
Fix #289

I intentionally didn't add it to the profile because the definition isn't final.

### DIFF
--- a/security-baseline/rule-types/github/osps-qa-03.yaml
+++ b/security-baseline/rule-types/github/osps-qa-03.yaml
@@ -1,0 +1,73 @@
+---
+version: v1
+release_phase: alpha
+type: rule-type
+name: osps-qa-03
+display_name: Package management file listing dependencies is present
+short_failure_message: No package management file listing dependencies was found
+severity:
+  value: info
+context:
+  provider: github
+description: |
+  This rule ensures that the repository provides a dependency list that accounts 
+  for the direct language dependencies when the package management system supports it.
+  It checks for the presence of a Gemfile.lock, go.sum, package-lock.json or 
+  Cargo.toml according to the package management system used.
+guidance: |
+  Ensure that the repository provides a dependency list that accounts for the direct
+  language dependencies in the form of a lockfile, for example Gemfile.lock, go.sum,
+  package-lock.json etc.
+  Ensure the lockfile is in the same directory as the package file.
+def:
+  in_entity: repository
+  rule_schema: {}
+  ingest:
+    type: git
+  eval:
+    type: rego
+    rego:
+      type: deny-by-default
+      def: |
+        package minder
+
+        import rego.v1
+
+        default allow := false
+        default skip := false
+        default message := "Cannot find lockfile in the same directory as the package file"
+        
+        package_manager_files := [
+          {"name": "Gemfile", "lockfiles": ["Gemfile.lock"]},
+          {"name": "go.mod", "lockfiles": ["go.sum"]},
+          {"name": "package.json", "lockfiles": ["package-lock.json", "yarn.lock"]},
+          {"name": "Cargo.toml", "lockfiles": ["Cargo.lock"]},
+        ]
+        
+        skip if {
+          # Skip if no package manager file exists
+          every package_manager in package_manager_files {
+            required_files := file.ls_glob(sprintf("./%s", [package_manager.name]))
+            count(required_files) == 0
+          }
+        }
+        
+        allow if {
+          # Ensure that we find the required file
+          some package_manager in package_manager_files
+          package_files := file.ls_glob(sprintf("./%s", [package_manager.name]))
+          count(package_files) > 0
+        
+          # Get the directory for the package file
+          some package_path in package_files
+          dir := trim_suffix(package_path, sprintf("/%s", [package_manager.name]))
+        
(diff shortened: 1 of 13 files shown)


security-baseline/rule-types/github/osps-qa-03.yaml
```diff
@@ -0,0 +1,73 @@
...
+          {"name": "Gemfile", "lockfiles": ["Gemfile.lock"]},
+          {"name": "go.mod", "lockfiles": ["go.sum"]},
+          {"name": "package.json", "lockfiles": ["package-lock.json", "yarn.lock"]},
+          {"name": "Cargo.toml", "lockfiles": ["Cargo.lock"]},
+        ]
+        
+        skip if {
+          # Skip if no package manager file exists
+          every package_manager in package_manager_files {
+            required_files := file.ls_glob(sprintf("./%s", [package_manager.name]))
```
eleftherias, line 50: Unfortunately the glob library we're using in Minder doesn't support `**` patterns, which is why this only looks at the root directory. This is probably something we want to change in the Minder code.

JAORMX (approved):
This rule looks handy overall. Can we get a copy in the common dir too? With regular verbiage

security-baseline/rule-types/github/osps-qa-03.yaml
evankanderson: Yeah -- we could also use `file.walk` for this:

```suggestion
            all_files := file.walk(".")
            package_manager_files = [ filepath |
              filepath := all_files[_];
              glob.match("**/" + package_manager[_].name, ["/"], filepath)
            ]
```

security-baseline/rule-types/github/osps-qa-03.yaml
```diff
@@ -0,0 +1,73 @@
...
+          package_files := file.ls_glob(sprintf("./%s", [package_manager.name]))
+          count(package_files) > 0
+        
+          # Get the directory for the package file
+          some package_path in package_files
+          dir := trim_suffix(package_path, sprintf("/%s", [package_manager.name]))
+        
+          # Ensure a lockfile exists for the required file in the same directory
+          lockfile_exists(dir, package_manager.lockfiles) 
+        }
```
evankanderson, line 67: Is this an `allow if` or a constraints if the lockfile is missing?  i.e. if I have:

```
./go.mod
./go.sum
./package.json
```

(No lockfile  for NPM)

Is this a "pass" or a "fail"?  I suspect it's a fail for the NPM ecosystem, even if it's a pass for the Go ecosystem.

evankanderson (commented):
This seems handy!

I'm not sure whether QA-03 will be loosened to allow e.g. `requirements.txt`, but I think we should keep this rule in the common area even if QA-03 becomes less strict.

security-baseline/rule-types/github/osps-qa-03.yaml
```diff
@@ -0,0 +1,73 @@
...
+        
+        package_manager_files := [
+          {"name": "Gemfile", "lockfiles": ["Gemfile.lock"]},
+          {"name": "go.mod", "lockfiles": ["go.sum"]},
+          {"name": "package.json", "lockfiles": ["package-lock.json", "yarn.lock"]},
+          {"name": "Cargo.toml", "lockfiles": ["Cargo.lock"]},
+        ]
+        
+        skip if {
+          # Skip if no package manager file exists
```
evankanderson, line 48: Looking at QA-03, I think if there's a language in the repo identified by GitHub, we should be *ensuring* that there are package files, rather than ignoring repos that also don't check in Gemfile/etc.

We could find the set of detected languages with a call to the repo config data source.  At least for npm and Go, I know that a `go.mod` / `package.json` is needed to declare a module at all, whether or not it has dependencies, so checking for those files makes sense.

eleftherias: Do you think we should be checking for `go.mod` instead of `go.sum`, or that we should be checking for both?

evankanderson: For Go, I think checking for both makes sense.  I'm assuming that doing the same for NPM won't be a big problem.  I'm not sure what the community conventions are for checking in e.g. Gemfile.lock -- I haven't really done much work in the Ruby community.

evankanderson:
I'm going to merge this and then clean up to the current OSPS baseline standard.

PR #398, opened 2026-05-19
Migrate rule policies to Rego v1

dakshhhhh16:
## Summary

This PR migrates the Rego policies in `minder-rules-and-profiles` to Rego v1 syntax as Phase 2 of the Rego v1 migration work from mindersec/minder#5262.
Child issue mindersec/minder#6463

It adds a reusable migration helper that extracts embedded YAML `def: |` policy blocks, runs `opa fmt --v0-v1`, and writes the formatted Rego back into place. The migration also updates docs/examples so downstream users have a repeatable path for their own rule type YAML.

## What changed

- Added `scripts/migrate-rego-v1.py` and a `task migrate-rego-v1` target.
- Migrated embedded Rego policies in rule types, security-baseline rule types, and autofill-insights rule types to `import rego.v1`.
- Removed `future.keywords` usage from policy docs and examples.
- Kept the migration script idempotent and whitespace-safe for YAML block scalars.

## Validation

- `python3 scripts/migrate-rego-v1.py --check .`
- `git diff --check`
- `go test ./...`

## Notes for reviewers

This should be a syntax-only migration. Any policy behavior change should be treated as unintended unless called out during review.

### DIFF
--- a/scripts/migrate-rego-v1.py
+++ b/scripts/migrate-rego-v1.py
@@ -1,0 +1,156 @@
+#!/usr/bin/env python3
+"""Migrate embedded Rego rule definitions in YAML files to Rego v1 syntax."""
+
+from __future__ import annotations
+
+import argparse
+import pathlib
+import re
+import subprocess
+import sys
+import tempfile
+
+
+DEF_BLOCK_RE = re.compile(r"^(?P<indent>\s*)def:\s*\|(?P<chomp>[+-]?)\s*$")
+
+
+def yaml_files(paths: list[pathlib.Path]) -> list[pathlib.Path]:
+    files: list[pathlib.Path] = []
+    for path in paths:
+        if path.is_dir():
+            files.extend(sorted(path.rglob("*.yaml")))
+            files.extend(sorted(path.rglob("*.yml")))
+        elif path.suffix in {".yaml", ".yml"}:
+            files.append(path)
+    return sorted(set(files))
+
+
+def leading_spaces(line: str) -> int:
+    return len(line) - len(line.lstrip(" "))
+
+
+def extract_block(lines: list[str], start: int, parent_indent: int) -> tuple[list[str], int]:
+    block: list[str] = []
+    current = start
+    while current < len(lines):
+        line = lines[current]
+        if line.strip() and leading_spaces(line) <= parent_indent:
+            break
+        block.append(line)
+        current += 1
+    return block, current
+
+
+def dedent_block(block: list[str]) -> tuple[str, int]:
+    nonblank_indents = [leading_spaces(line) for line in block if line.strip()]
+    if not nonblank_indents:
+        return "", 0
+
+    block_indent = min(nonblank_indents)
+    source = "".join(
+        line[block_indent:] if len(line) >= block_indent else "\n"
+        for line in block
+    )
+    return source, block_indent
+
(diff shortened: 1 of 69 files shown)


dakshhhhh16:
@evankanderson 
You can have a look into this. It will lead to close of issue #6463 and I can further start working on phase 3

evankanderson:
I need to look into why the tests are failing in this repo.

evankanderson:
#402 should fix

dakshhhhh16:
@evankanderson @JAORMX 
What's your take on this one? After the merge of this I'll start working on the next phase

evankanderson:
> @evankanderson @JAORMX 
> What's your take on this one? After the merge of this I'll start working on the next phase

Whoops, I lost the "edge trigger" for this one.  I'll look at it today.

scripts/migrate-rego-v1.py
```diff
@@ -0,0 +1,156 @@
...
+        elif path.suffix in {".yaml", ".yml"}:
+            files.append(path)
+    return sorted(set(files))
+
+
+def leading_spaces(line: str) -> int:
+    return len(line) - len(line.lstrip(" "))
+
+
+def extract_block(lines: list[str], start: int, parent_indent: int) -> tuple[list[str], int]:
```
evankanderson, line 32: Why is this not using the Python `yaml` module (PyYAML)?

Alternatively, it may be better to do something with `yq` (which preserves comments, I'm not sure if PyYAAML does):

```bash
yq .def.eval.rego.def $YAML | opa fmt --v0-v1 - > /tmp/formatted
yq -i '.def.eval.rego.def = load_str(/tmp/formatted)' $YAML
```

evankanderson: Alternatively, we could just have this as a one-time thing, and not commit it to the repo.

evankanderson (commented):
This looks great, except that I have some questions about the script processing the YAML files as text using regular expressions.  😁

scripts/migrate-rego-v1.py
dakshhhhh16: @evankanderson 
I went with the small Python helper mostly to keep the migration focused on the embedded Rego blocks without rewriting the rest of each YAML file.

I avoided PyYAML because loading and dumping the files would likely reformat unrelated YAML and may drop comments. `yq` could work, and I like that it preserves comments, but it would add another tool dependency on top of `opa`.

The script is intentionally narrow, it only looks at literal `def: |` blocks, and only runs `opa fmt --v0-v1` when the block looks like Rego by starting with `package`.

If you want this to stay as a one-time migration helper, I can drop the script/task/README bits and keep only the generated Rego v1 changes in the PR.

evankanderson: I'll merge this as-is, but I think we'll probably want to remove this scaffolding sooner rather than later.

evankanderson (approved):

scripts/migrate-rego-v1.py
dakshhhhh16: Will keep this in mind

dakshhhhh16:
@evankanderson Should I start working on the next phase?

evankanderson:
Feel free to put the next phase behind a feature flag, yes.

PR #401, opened 2026-05-31
feat: add OSPS security baseline rule types for GitLab

intelligent-ears:
Port 6 OSPS Security Baseline level-1 rules to GitLab
Adds GitLab versions of 6 OSPS Baseline level-1 rules that use git ingest (clone-based file checks with no GitHub API calls). These rules work identically with GitLab repositories — the only change from the GitHub versions is context: provider: gitlab.
Rules added:

osps-br-03-02: Deliver releases via encrypted channels
osps-gv-03-01: Enforce CONTRIBUTING file presence
osps-le-03-01: LICENSE or COPYING files are available
osps-qa-02-01: Source code contains direct dependency list
osps-qa-05-01: No generated executable artifacts in VCS
osps-qa-05-02: No unreviewable binary artifacts in VCS

Part of the GitLab provider coverage work tracked in mindersec/minder#6435.
The remaining 15 level-1 rules use GitHub REST API endpoints and will need separate GitLab API equivalents in follow-up PRs.

### DIFF
--- a/security-baseline/rule-types/gitlab/osps-br-03-02.yaml
+++ b/security-baseline/rule-types/gitlab/osps-br-03-02.yaml
@@ -1,0 +1,47 @@
+version: v1
+release_phase: alpha
+type: rule-type
+name: osps-br-03-02 
+display_name: Deliver releases via encrypted channels
+short_failure_message: Releases available via unsecured channels
+severity:
+  value: high
+context:
+  provider: gitlab
+description: |
+  This check determines whether the release artifacts of the project
+  are delived via unsecured (unencrypted) channels.
+
+  Downloading release artifacts via unsecured channels is a supply
+  chain risk for consumers, as attackers could compromise the release
+  assets in-transit via a Machine-in-the-Middle (MitM) attack.
+
+guidance: |
+  Deliver artifacts using HTTPS URLs or package repositories.
+def:
+  in_entity: repository
+  rule_schema: {}
+  ingest:
+    type: git
+  eval:
+    # We only need to worry about links in e.g. the README, as artifacts
+    # distributed via GitHub releases or package repositories will already
+    # be encrypted.
+    type: rego
+    rego:
+      type: constraints
+      def: |
+        package minder
+        import rego.v1
+
+        # Currently, we assume that direct downloads are linked from a README
+        # file, and not elsewhere in the documentation.
+        readme_contents := file.read("./README.md")
+        http_urls := regex.find_all_string_submatch_n(`(http://\S*)`, readme_contents, -1)[0]
+
+        violations contains {"msg": msg} if {
+          some url in http_urls
+          not startswith(url, "http://localhost")
+
+          msg := sprintf("README.md contains non-HTTPS URL '%s'", [url])
+        }
--- a/security-baseline/rule-types/gitlab/osps-gv-03-01.yaml
+++ b/security-baseline/rule-types/gitlab/osps-gv-03-01.yaml
@@ -25,24 +25,11 @@ def:
       type: deny-by-default
       def: |
         package minder
-
-        import future.keywords.every
-        import future.keywords.if
-
+        import rego.v1
         default allow := false
-
         allow if {
           files := file.ls_glob("./CONTRIBUTING*")
-
           some name
           content := file.read(files[name])
-          "" != content
-        }
-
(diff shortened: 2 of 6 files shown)


security-baseline/rule-types/gitlab/osps-br-03-02.yaml
```diff
@@ -0,0 +1,47 @@
+version: v1
+release_phase: alpha
+type: rule-type
+name: osps-br-03-02 
+display_name: Deliver releases via encrypted channels
+short_failure_message: Releases available via unsecured channels
+severity:
+  value: high
+context:
+  provider: gitlab
```
evankanderson, line 10: Is this the _same_ rule as for `github`, but with the provider changed?

I'm trying to think if there's a better way to express this.

security-baseline/rule-types/gitlab/osps-gv-03-01.yaml
```diff
@@ -0,0 +1,48 @@
...
+    type: git
+  eval:
+    type: rego
+    rego:
+      type: deny-by-default
+      def: |
+        package minder
+
+        import future.keywords.every
+        import future.keywords.if
```
evankanderson: Pefer:


```suggestion
        import rego.v1
```

See also #398

security-baseline/rule-types/gitlab/osps-gv-03-01.yaml
```diff
@@ -0,0 +1,48 @@
...
+          "" != content
+        }
+
+        allow if {
+          files := file.ls_glob("./CONTRIBUTING/*")
+
+          some name
+          content := file.read(files[name])
+          "" != content
+        }
```
evankanderson: Why is this duplicated with 34-40?

security-baseline/rule-types/gitlab/osps-le-03-01.yaml
```diff
@@ -0,0 +1,56 @@
...
+          "" != content
+        }
+
+        allow if {
+          files := file.ls_glob("./LICENSE/*")
+
+          some name
+          content := file.read(files[name])
+          "" != content
+        }
```
evankanderson: Weird, this also duplicates 34-40

security-baseline/rule-types/gitlab/osps-le-03-01.yaml
```diff
@@ -0,0 +1,56 @@
...
+  eval:
+    type: rego
+    rego:
+      type: deny-by-default
+      def: |
+        package minder
+
+        import future.keywords.every
+        import future.keywords.if
+
```
evankanderson:
```suggestion
        import rego.v1

```

evankanderson (commented):
A few comments; I'd also like to wait for #402, even though these don't have tests.

My main concern is that it seems like we're needing to duplicate identical rules for github and gitlab.  Maybe the `context.provider` should be a [list of `supported_provider_types` that the provider needs to implement for this rule](https://github.com/mindersec/minder/blob/main/proto/minder/v1/minder.proto#L3663)?

security-baseline/rule-types/gitlab/osps-br-03-02.yaml
intelligent-ears: Yes, exactly — it's identical except for context.provider: gitlab. The Rego logic, ingest type, and everything else is the same. I think the right long-term fix is to allow context.provider to accept a list (e.g. provider: [github, gitlab]) so provider-agnostic rules don't need to be duplicated. I can open a separate issue in minder to track that, or handle it here if you'd prefer.

evankanderson: Open an issue in Minder, and link to this PR for an example of how the current behavior is not great.

intelligent-ears: https://github.com/mindersec/minder/issues/6481
Opened the issue

intelligent-ears:
Closing in favor of the approach discussed in https://github.com/mindersec/minder/issues/6481 — since `context.provider` is already ignored, the right fix is to remove the `context` field from the provider-agnostic baseline rules rather than duplicating them. Will follow up with a separate PR that removes `context` from the git-ingest based rules across the existing GitHub baseline.

PR #402, opened 2026-06-01
Fix opencontainers/runtime-spec breakage

evankanderson:
It turns out that the opencontainers runtime-spec did a go-breaking but not wire-protocol-breaking change in 1.3.0; when we hit this in Minder, we did the same `require` trick.  At some point, _maybe_ opencontainers will fix this, but it sounded like they were basically stuck with breaking someone.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -309,3 +309,7 @@ require (
 	sigs.k8s.io/yaml v1.6.0 // indirect
 	www.velocidex.com/golang/regparser v0.0.0-20250203141505-31e704a67ef7 // indirect
 )
+
+// Pin this to 1.2.1 until using containerd/v2; 1.3.0 has a backwards incompatible change
+// https://github.com/opencontainers/runtime-spec/pull/1279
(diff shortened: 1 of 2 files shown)


evankanderson:
If our tests were simply a binary we downloaded from `mindersec/minder`, we wouldn't need to maintain this enormous `go.mod`.  (I'll get @JAORMX to review as well, but figured that up-and-coming maintainers might want to see what goes on first.)

dakshhhhh16 (approved):

JAORMX (approved):

JAORMX:
@evankanderson I'm in for the binary download approach. which, I guess, means moving the code to minder. That's fine. It was just easy to do at the time. I dont have a strong opinion about this

evankanderson:
We're going to publish a binary (and probably a GitHub action) for running rule tests as part of https://github.com/mindersec/minder/issue/6434, but I expect that to be a couple months out still (draft design should land late this week or early next week for comment).

PR #405, opened 2026-06-13
fix: remove context.provider from git-ingest OSPS baseline rules

intelligent-ears:
Since `context.provider` is ignored by Minder as of mindersec/minder#2843, having it set to `github` on rules that use only `git` ingest gives a false impression that these rules are GitHub-specific.
This removes the `context` block from 6 provider-agnostic level-1 rules that use git clone ingest with no GitHub API calls:

- `osps-br-03-02`: Deliver releases via encrypted channels
- `osps-gv-03-01`: Enforce CONTRIBUTING file presence
- `osps-le-03-01`: LICENSE or COPYING files are available
- `osps-qa-02-01`: Source code contains direct dependency list
- `osps-qa-05-01`: No generated executable artifacts in VCS
- `osps-qa-05-02`: No unreviewable binary artifacts in VCS

REST-based rules that call GitHub-specific API endpoints retain their context block to document the provider dependency.
Supersedes #401. Refs mindersec/minder#6481.

### DIFF
--- a/security-baseline/rule-types/github/osps-qa-05-02.yaml
+++ b/security-baseline/rule-types/github/osps-qa-05-02.yaml
@@ -7,9 +7,6 @@ display_name: Version control does not contain binary artifacts
 short_failure_message: Binary artifacts found in the repository
 severity:
   value: high
-context:
-  provider: github
-description: |
   This rule ensures that the repository does not contain any unreviewable
   binary artifacts.  It is difficult to review and ascertain the provenance
   of binary artifacts such as application binaries or compressed archives.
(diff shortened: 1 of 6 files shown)


security-baseline/rule-types/github/osps-qa-05-02.yaml
```diff
@@ -7,9 +7,6 @@ display_name: Version control does not contain binary artifacts
 short_failure_message: Binary artifacts found in the repository
 severity:
   value: high
-context:
-  provider: github
-description: |
```
evankanderson, line 12: You still have `description` removed from several files.

JAORMX (approved):

PR #408, opened 2026-06-28
ci: enforce Rego v1 policies with Regal

### LINKED ISSUE #409: Enforce Rego V1 policies with Regal in CI
Part of mindersec/minder#5262

## Problem

The official rule policies have been migrated to Rego V1, but CI does not currently prevent V0 syntax or formatting regressions from being introduced.

Because the policies are embedded inside YAML files, they must be extracted before Regal can lint them.

## Proposed changes

- Add `.regal/config.yaml`.
- Enable the following rules as errors:
  - `use-rego-v1`
  - `opa-fmt`
- Add a reusable script that extracts embedded Rego policies from rule-type YAML.
- Check that embedded policies remain idempotent with `opa fmt --v0-v1`.
- Run Regal against policies from:
  - `rule-types`
  - `security-baseline/rule-types`
  - `autofill-insights/rule-types`
- Pin the OPA and Regal versions used by CI.
- Limit the gate to migration-related rules to avoid unrelated style cleanup.

## Acceptance criteria

- Every official embedded policy is checked by Regal.
- Policies without Rego V1 support fail CI.
- Policies that regress from the expected OPA formatting fail CI.
- Existing policies pass the new CI gate.
- The extraction step fails when no policies are found.
- The repository’s existing test suite remains green.

dakshhhhh16:
Part of mindersec/minder#5262
Resolves #409 
## Summary

Adds a CI guardrail that keeps official rule policies on Rego v1 and prevents formatting regressions.

## What changed

- Added a repository-level Regal configuration for `use-rego-v1` and `opa-fmt`.
- Added a small extraction utility for Rego policies embedded in rule-type YAML.
- Check that embedded policies remain idempotent with `opa fmt --v0-v1`.
- Run pinned OPA and Regal versions in CI.
- Lint all 65 official policies across the main, security-baseline, and autofill rule directories.

The Regal job is intentionally limited to the two migration rules, so this PR does not introduce unrelated style cleanup.

## Testing

```sh
python3 scripts/migrate-rego-v1.py --check .
python3 scripts/extract-rego.py \
  --output /tmp/minder-rego \
  rule-types security-baseline/rule-types autofill-insights/rule-types
find /tmp/minder-rego -name '*.rego' -exec opa fmt --write {} +
go run github.com/open-policy-agent/regal@v0.41.1 lint \
  --disable-all \
  --enable use-rego-v1 \
  --enable opa-fmt \
  --config-file .regal/config.yaml \
  /tmp/minder-rego
go test ./...
```

### DIFF
--- a/.regal/config.yaml
+++ b/.regal/config.yaml
@@ -9,4 +9,4 @@ rules:
 capabilities:
   from:
     engine: opa
-    version: v0.68.0
+    version: v1.15.2
--- a/.regal/config.yaml
+++ b/.regal/config.yaml
@@ -1,0 +1,12 @@
+rules:
+  imports:
+    use-rego-v1:
+      level: error
+  style:
+    opa-fmt:
+      level: error
+
+capabilities:
+  from:
+    engine: opa
+    version: v0.68.0
--- a/.regal/config.yaml
+++ b/.regal/config.yaml
@@ -9,4 +9,4 @@ rules:
 capabilities:
   from:
     engine: opa
-    version: v1.15.2
+    version: v0.68.0
--- a/scripts/extract-rego.py
+++ b/scripts/extract-rego.py
@@ -1,0 +1,103 @@
+#!/usr/bin/env python3
+"""Extract Rego policies embedded in YAML rule type definitions."""
+
+from __future__ import annotations
+
+import argparse
+import pathlib
+import re
+
+
+DEF_BLOCK_RE = re.compile(r"^(?P<indent>\s*)def:\s*\|[+-]?\s*$")
+
+
+def yaml_files(paths: list[pathlib.Path]) -> list[pathlib.Path]:
+    files: list[pathlib.Path] = []
+    for path in paths:
+        if path.is_dir():
+            files.extend(path.rglob("*.yaml"))
+            files.extend(path.rglob("*.yml"))
+        elif path.suffix in {".yaml", ".yml"}:
+            files.append(path)
+    return sorted(set(files))
+
+
+def leading_spaces(line: str) -> int:
+    return len(line) - len(line.lstrip(" "))
+
+
+def extract_block(lines: list[str], start: int, parent_indent: int) -> tuple[str, int]:
+    block: list[str] = []
+    current = start
+    while current < len(lines):
+        line = lines[current]
+        if line.strip() and leading_spaces(line) <= parent_indent:
+            break
+        block.append(line)
+        current += 1
+
+    nonblank_indents = [leading_spaces(line) for line in block if line.strip()]
+    if not nonblank_indents:
+        return "", current
+
+    block_indent = min(nonblank_indents)
+    source = "".join(
+        line[block_indent:] if len(line) >= block_indent else "\n"
+        for line in block
+    )
+    return source, current
+
+
+def extract_policies(path: pathlib.Path) -> list[str]:
+    lines = path.read_text().splitlines(True)
+    policies: list[str] = []
+    index = 0
+
+    while index < len(lines):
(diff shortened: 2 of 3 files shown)


.regal/config.yaml
```diff
@@ -0,0 +1,12 @@
...
+    use-rego-v1:
+      level: error
+  style:
+    opa-fmt:
+      level: error
+
+capabilities:
+  from:
+    engine: opa
+    version: v0.68.0
```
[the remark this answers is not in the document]
dakshhhhh16: not required

evankanderson (commented):
You may want to take a look at https://github.com/mindersec/minder/pull/6558; that PR would allow us to move all the rego files to be `.rego` with comments for the rest of the RuleType definition, rather than embedded rego inside YAML.

scripts/extract-rego.py
evankanderson, line 1: It feels like you're extracting a single (well-known) YAML field from a document here -- this should almost certainly use pyyaml rather than string processing.  (Or you could use `yq` and some shell, but I'm okay with using Python)

dakshhhhh16:
> You may want to take a look at [mindersec/minder#6558](https://github.com/mindersec/minder/pull/6558); that PR would allow us to move all the rego files to be `.rego` with comments for the rest of the RuleType definition, rather than embedded rego inside YAML.

Thanks for pointing me to mindersec/minder#6558. It looks like that approach could remove the need to extract Rego from YAML before linting. Would you prefer that I wait for mindersec/minder#6558 and update this PR to lint the .rego files directly, or keep the current approach as a temporary check for the existing YAML format?

scripts/extract-rego.py
dakshhhhh16: I personally like python more, can change it if it's necessary.

evankanderson: I'm happy to keep using python, but I'd like to use the yaml library for parsing, rather than ad-hoc parsing with regular expressions and python code.

evankanderson (commented):
I think it's fine to merge this now, when the rego change merges we can hopefully simplify a lot of this.  (And seeing this pile higher motivated me to go in and try to fix things at the root.)

scripts/extract-rego.py
dakshhhhh16: Will keep this in mind

dakshhhhh16:
@evankanderson 
Can you pls merge this one so that i can proceed forward?

scripts/extract-rego.py
```diff
@@ -0,0 +1,103 @@
...
+
+        source, index = extract_block(
+            lines,
+            index + 1,
+            len(match.group("indent")),
+        )
+        if source.lstrip().startswith("package "):
+            policies.append(source)
+
+    return policies
```
evankanderson: If you use PyYAML, this is:

```suggestion
    contents = yaml.load(path.read_text())
    rego = contents.get('def', {}).get('eval', {}),.get('rego', {}).get('def', "")
    if rego != "":
      return [rego]
    return []
```

And you don't need `extract_block`, etc.

evankanderson (commented):
This should use the standard YAML module for the extraction.  (It's fine to use python rather than e.g. `yq .def.eval.rego.def $filename`), but let's not roll our own extraction code when the library handles that case just fine.)

scripts/extract-rego.py
dakshhhhh16: Done, it now reads def.eval.rego.def with safe_load, and the old extraction helpers are removed.

dakshhhhh16:
@evankanderson 
Apologies for a late revert, I was a bit busy from the past few days, I have made the changes, you can have a look now!

evankanderson (approved):

PR #412, opened 2026-07-01
feat: add GitLab provider rule types for OSPS level-1 security baseline

intelligent-ears:
## Summary

Ports 5 of 11 REST-based OSPS level-1 rules to the GitLab provider, adding
a new `security-baseline/rule-types/gitlab/` directory mirroring the existing
`github/` structure.

## Rules added

| Rule ID | Description | GitLab API field |
|---------|-------------|-----------------|
| osps-ac-02-01 | Default collaborators to lowest privileges | `visibility == "public"` |
| osps-br-03-01 | Development resources use secure channels | `http_url_to_repo` starts with `https://` |
| osps-qa-01-01 | Source code is publicly readable | `visibility == "public"` + `http_url_to_repo` present |
| osps-do-02-01 | Public feedback mechanism supported | `issues_enabled` |
| osps-gv-02-01 | Project has mechanisms for public discussion | `issues_enabled` |

All rules ingest from `GET /projects/{{.Entity.RepoId}}` and were validated
end-to-end against a live GitLab project using a local Minder stack.

## Notes

- GitLab has a third visibility tier (`internal`) not present in GitHub. Rules
  treating `internal` the same as `private` is documented in guidance text.
- GitLab has no "Discussions" feature equivalent to GitHub — `osps-do-02-01`
  and `osps-gv-02-01` use `issues_enabled` as the proxy, with this gap noted
  in guidance.

## Remaining rules (follow-up PRs)

6 rules require deeper GitLab API work:
- `osps-ac-03-01`, `osps-ac-03-02`, `osps-qa-01-02` — branch protection (Protected Branches API)
- `osps-le-02-01` — license check (requires `?license=true` param)
- `osps-le-02-02`, `osps-le-03-02` — release asset checks

Fixes part of https://github.com/mindersec/minder/issues/6435

### DIFF
--- a/security-baseline/rule-types/gitlab/osps-ac-02-01.yaml
+++ b/security-baseline/rule-types/gitlab/osps-ac-02-01.yaml
@@ -1,0 +1,39 @@
+version: v1
+release_phase: alpha
+type: rule-type
+name: osps-ac-02-01
+display_name: Default collaborators to lowest privileges
+short_failure_message: Collaborators default to privileged access
+severity:
+  value: info
+context:
+  provider: gitlab
+description: Verifies that project collaborators default to the lowest access level.
+guidance: |
+  Reduce the risk of unauthorized access to the project's repository
+  by limiting the permissions granted to collaborators.
+  Most public version control systems (such as GitLab) are configured
+  in this manner.
+
+  Note: GitLab's "internal" visibility (visible to all authenticated
+  instance users) is treated the same as "private" by this rule, since
+  it still restricts access beyond the general public.
+def:
+  in_entity: repository
+  rule_schema: {}
+  ingest:
+    type: rest
+    rest:
+      endpoint: '/projects/{{.Entity.RepoId}}'
+      parse: json
+      fallback:
+        - http_code: 404
+          body: |
+            {"http_status": 404, "message": "Repo not found"}
+  eval:
+    type: jq
+    jq:
+      # Public projects on GitLab will already have minimum _default_ permissions.
+      - ingested:
+          def: ".visibility"
+        constant: "public"
--- a/security-baseline/rule-types/gitlab/osps-ac-02-01.yaml
+++ b/security-baseline/rule-types/gitlab/osps-ac-02-01.yaml
@@ -6,8 +6,6 @@ display_name: Default collaborators to lowest privileges
 short_failure_message: Collaborators default to privileged access
 severity:
   value: info
-context:
-  provider: gitlab
 description: Verifies that project collaborators default to the lowest access level.
 guidance: |
   Reduce the risk of unauthorized access to the project's repository
--- a/security-baseline/rule-types/gitlab/osps-br-03-01.yaml
+++ b/security-baseline/rule-types/gitlab/osps-br-03-01.yaml
@@ -6,8 +6,6 @@ display_name: Development resources use secure channels
 short_failure_message: Insecure access to VCS
 severity:
   value: info
-context:
-  provider: gitlab
 description: Verifies that websites and version control systems for development use secure channels.
 guidance: |
   Any websites and version control systems involved in the project
--- a/security-baseline/rule-types/gitlab/osps-br-03-01.yaml
+++ b/security-baseline/rule-types/gitlab/osps-br-03-01.yaml
@@ -1,0 +1,41 @@
+version: v1
+release_phase: alpha
+type: rule-type
+name: osps-br-03-01
+display_name: Development resources use secure channels
+short_failure_message: Insecure access to VCS
+severity:
+  value: info
+context:
+  provider: gitlab
+description: Verifies that websites and version control systems for development use secure channels.
+guidance: |
+  Any websites and version control systems involved in the project
+  development MUST be delivered using SSH, HTTPS, or other encrypted
+  channels.
+
+  GitLab does this by default.
+def:
+  in_entity: repository
+  rule_schema: {}
+  ingest:
+    type: rest
+    rest:
+      endpoint: '/projects/{{.Entity.RepoId}}'
+      parse: json
+      fallback:
+        - http_code: 404
+          body: |
+            {"http_status": 404, "message": "Repo not found"}
+  eval:
+    type: rego
+    rego:
+      type: deny-by-default
+      def: |
+        package minder
+
+        import rego.v1
+
+        default allow := false
+
+        allow if startswith(input.ingested.http_url_to_repo, "https://")
--- a/security-baseline/rule-types/gitlab/osps-do-02-01.yaml
+++ b/security-baseline/rule-types/gitlab/osps-do-02-01.yaml
@@ -6,8 +6,6 @@ display_name: Public feedback mechanism supported
 short_failure_message: No public feedback mechanism found
 severity:
   value: medium
-context:
-  provider: gitlab
 description: |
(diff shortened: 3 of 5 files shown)


security-baseline/rule-types/gitlab/osps-ac-02-01.yaml
```diff
@@ -0,0 +1,39 @@
+version: v1
+release_phase: alpha
+type: rule-type
+name: osps-ac-02-01
+display_name: Default collaborators to lowest privileges
+short_failure_message: Collaborators default to privileged access
+severity:
+  value: info
+context:
+  provider: gitlab
```
evankanderson: Interestingly, this doesn't actually _do_ anything, so it's probably best to leave it off.

evankanderson (commented):
We need to come up with some sort of plan for being able to load these rules alongside the GitHub rules in a single project for organizations managing both GitHub and GitLab projects.  (For example, Eclipse Foundation has this need.)

We may also want to think about some way to skip repositories that don't have the correct provider / rest API shape.  This feels like it should be part of the top-level `def`, possibly "requires_implements" (though a better name is probably needed).

security-baseline/rule-types/gitlab/osps-ac-02-01.yaml
intelligent-ears: I'll remove these.

intelligent-ears:
> We need to come up with some sort of plan for being able to load these rules alongside the GitHub rules in a single project for organizations managing both GitHub and GitLab projects. (For example, Eclipse Foundation has this need.)
> 
> We may also want to think about some way to skip repositories that don't have the correct provider / rest API shape. This feels like it should be part of the top-level `def`, possibly "requires_implements" (though a better name is probably needed).

On the broader question: would the preferred direction be a single unified rule set that handles both GitHub and GitLab in the same rule file, rather than parallel provider-specific directories?

The main challenge is that GitHub and GitLab diverge at both the endpoint and field level:


Concept | GitHub endpoint | GitLab endpoint
-- | -- | --
Repo metadata | `/repos/{{.Entity.Owner}}/{{.Entity.Name}}` | `/projects/{{.Entity.RepoId}}`
Branch protection | `/repos/{owner}/{repo}/branches/{branch}/protection` | `/projects/{id}/protected_branches/{branch}`
Release assets | `/repos/{owner}/{repo}/releases/{id}/assets` | `/projects/{id}/releases/{tag}/assets/links`

Field names differ too — `has_issues` vs `issues_enabled`, `clone_url` vs `http_url_to_repo`, `spdx_id` vs `key` for licenses.

The `requires_implements` idea feels like the right long-term direction. A strawman for what it might look like in the rule schema:
```yaml
  requires_implements:
    - rest_api: github   # skip this rule for non-GitHub providers
  in_entity: repository
  ...
```

or more capability-based rather than provider-named:

```yaml
  requires_implements:
    - capability: github_rest_api
 ```
The capability-based form seems more extensible — it decouples the rule from knowing provider names and instead declares what API shape it depends on. The engine would then skip evaluation (with a `skipped` status rather than `error` or `fail`) for repos whose provider doesn't advertise that capability.

This would let us keep `osps-ac-02-01` as a single canonical rule but have two ingest/eval variants gated by capability, or keep separate rule files that are just skipped cleanly when the provider doesn't match.

Is the intent that `requires_implements` gates the whole rule (skip if not supported), or that it could eventually allow multiple ingest blocks — one per capability — within the same rule file? The latter would be more powerful but a bigger schema change.
Happy to draft a more formal design doc or issue for this if it would help move the discussion forward.

evankanderson:
> > We need to come up with some sort of plan for being able to load these rules alongside the GitHub rules in a single project for organizations managing both GitHub and GitLab projects. (For example, Eclipse Foundation has this need.)
> > We may also want to think about some way to skip repositories that don't have the correct provider / rest API shape. This feels like it should be part of the top-level `def`, possibly "requires_implements" (though a better name is probably needed).
> 
> On the broader question: would the preferred direction be a single unified rule set that handles both GitHub and GitLab in the same rule file, rather than parallel provider-specific directories?

That's a good question!  I don't know, but we can figure it out together.  😁

Given the challenge you laid out around differences in both API URLs and content, I suspect that attempting to produce a unified ruletype that covers both GitHub and GitLab will be challenging (and adding other git forges will continue to add pain).  Remediation actions will also probably differ by provider, which again pushes in the direction of either using a separate ruletype for each provider or building some sort of "tree" of rules within a ruletype (ugh!).

Borrowing from our friends over at [Gemara](https://gemara.openssf.org/model/06-sensitive-activities.html), ruletypes implement specific policies to ensure that controls are in place.  Presumably, it will be the user's responsibility to ensure that control coverage is consistent across all entities.  At the same time, we probably want to make it _easier_ for users to verify Minder-implemented controls across different entities.

That is to say, I don't see a _good_ solution for implementing a single ruletype which covers multiple providers.  In term of mapping policies to controls, this may be as simple as adding a `-github` or `-gitlab` suffix to the rules, and then having compliance tools consuming Minder output do prefix-match checks.

> The `requires_implements` idea feels like the right long-term direction. A strawman for what it might look like in the rule schema:
> 
> ```yaml
>   requires_implements:
>     - rest_api: github   # skip this rule for non-GitHub providers
>   in_entity: repository
>   ...
> ```
> 
> or more capability-based rather than provider-named:
> 
> ```yaml
>   requires_implements:
>     - capability: github_rest_api
> ```
> 
> The capability-based form seems more extensible — it decouples the rule from knowing provider names and instead declares what API shape it depends on. The engine would then skip evaluation (with a `skipped` status rather than `error` or `fail`) for repos whose provider doesn't advertise that capability.

In the provider interface, the capabilities form is exposed as the "ProviderType" repeated enum.  You could match the GitHub vs GitLab REST APIs by having separate ProviderTypes for each (we already have a "github" type).  I like to call these "facets" of the provider interface, since they are optional additional methods which you can see when you list providers.  Since a provider can implement multiple facets, you could have some rules that require the "git" facet for filesystem-based checks, while others may require the "github" or "gitlab" facet for provider-specific checks.

> This would let us keep `osps-ac-02-01` as a single canonical rule but have two ingest/eval variants gated by capability, or keep separate rule files that are just skipped cleanly when the provider doesn't match.

The "single canonical rule" has the problem of needing to bundle evaluation and remediation variants in a single ruletype (which seems complex to me for low benefit).  On the other hand, if we use separate ruletypes for each provider variant, they'll need to have unique names so that they can be loaded at the same time.

> Is the intent that `requires_implements` gates the whole rule (skip if not supported), or that it could eventually allow multiple ingest blocks — one per capability — within the same rule file? The latter would be more powerful but a bigger schema change. Happy to draft a more formal design doc or issue for this if it would help move the discussion forward.

I think a _short_ design doc (probably covering the above points plus others that you think of) would be a good idea -- we're starting to collect those in https://github.com/mindersec/minder/tree/main/docs/docs/developer_guide/designs

security-baseline/rule-types/gitlab/osps-ac-02-01.yaml
evankanderson: (They are in all 5 files)

evankanderson:
In the meantime, if you want to remove the `context: {provider: gitlab}` blocks, I'm happy to merge this and then adjust later.

evankanderson (approved):

PR #418, opened 2026-07-17
test: migrate legacy tests to Starlark testing framework

krrish175-byte:
This PR completes the migration to the new `mindev` Starlark testing framework.

**Overview of Changes**
- All legacy `.test.yaml` files have been replaced with their Starlark `_test.star` equivalents.
- The legacy `rules_test.go` file and all Go module dependencies have been removed.
- CI pipeline and Taskfiles now use `mindev test rule-types/**/*_test.star` directly instead of `go test`.


.github/workflows/test.yml
```diff
@@ -20,17 +20,26 @@ jobs:
           go-version-file: 'go.mod'
           check-latest: true
 
+      - name: Checkout minder
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v4
+        with:
+          # TODO: Switch back to mindersec/minder once the testing framework
+          # changes (project context fix) are merged upstream.
+          repository: krrish175-byte/minder
+          path: minder
```
evankanderson: We should actually fetch the released binary, rather than checkout/build.  It will be faster and use less actions minutes.

.github/workflows/test.yml
```diff
@@ -20,17 +20,26 @@ jobs:
...
+
+      - name: Build mindev
+        run: |
+          cd minder
+          go build -o ../mindev ./cmd/dev
+          cd ..
+          echo "$(pwd)" >> $GITHUB_PATH
+
```
evankanderson: This should use your new action (which means I need to fix the 0.2.0 release)

rule-types/common/enforce_file_test.star
```diff
@@ -0,0 +1,67 @@
+def test_file_should_be_present():
+    res = eval(
+        rule="enforce_file",
+        entity={"type": "repository", "default_branch": "main"},
+        profile={"file": "README", "content": ""},
+        mock_fs={
+            "README": read_file("enforce_file.testdata/file_present/README")
+        }
```
evankanderson: Want to extract these out into a few constants, since we use the same mock_fs in a few different tests.

rule-types/common/enforce_file_test.star
```diff
@@ -0,0 +1,67 @@
...
+    )
+    assert.eq(res["status"], "pass")
```
evankanderson: What do you think about adding a function (macro) `PASS(...)` that takes an eval result and checks that the eval result is the string "pass"?  Usage would be:


```suggestion
    PASS(eval(
        rule="enforce_file",
        entity={"type": "repository", "default_branch": "main"},
        profile={"file": "README", "content": ""},
        mock_fs={
            "README": read_file("enforce_file.testdata/file_present/README")
        }
    ))
```

The idea being that we could simplify the common case and encode the "pass" and "status" constants.

evankanderson: I'm not sure whether this is a big enough win to be worth the indirect through the `PASS` macro, so I'm really looking for a second opinion.

evankanderson: If we want to be really radical, we could also bundle up a lot of the common setup:


```suggestion
    PASS(enforce_file(content="", fs=file_present)
```

I _think_ including the setup, this would be about 41 lines -- about 1-2 lines shorter than the YAML if it used references.

rule-types/common/python_lock_file_exists_test.star
```diff
@@ -0,0 +1,70 @@
+def test_project_contains_pipfile_lock():
+    res = eval(
+        rule="python_lock_file_exists",
+        entity={"owner": "coolhead", "name": "haze-wave", "type": "repository", "default_branch": "main"},
```
evankanderson: Like the YAML, we should extract this into a constant.

evankanderson: I'm noticing that the YAML has a separate level for the type and the properties, but we've combined them.  Thoughts?

rule-types/common/python_lock_file_exists_test.star
```diff
@@ -0,0 +1,70 @@
...
+        profile={},
+        mock_fs={
+            "Pipfile": read_file("python_lock_file_exists.testdata/pipfile_lock/Pipfile"),
+            "Pipfile.lock": read_file("python_lock_file_exists.testdata/pipfile_lock/Pipfile.lock")
```
evankanderson, line 15: Can we use a `txtar` archive for these files?  Having them as "live" dependency files makes dependabot and trivy angry during scans.

rule-types/github/branch_protection_allow_deletions_test.star
```diff
@@ -0,0 +1,46 @@
+def test_force_push_not_allowed():
+    res = eval(
+        rule="branch_protection_allow_deletions",
+        entity={"owner": "mindersec", "name": "minder", "type": "repository", "default_branch": "main"},
+        profile={},
+        mock_http={
+            "/repos/mindersec/minder/branches/main/protection": body("{\"allow_deletions\": {\"enabled\": false}}")
```
evankanderson: Rather than needing to escape quotes, what about:


```suggestion
            "/repos/mindersec/minder/branches/main/protection": body(str({"allow_deletions": {"enabled": false}}))
```

evankanderson: Also, we should extract the URL path into a constant or a function to build these.

```python
def protection(content):
    return {"/repos/mindersec/minder/branches/main/protection": content}

# ...
        mock_http=protection(body(str({"allow_deletions": {"enabled": true}})))
        # or
        mock_http=protection(body("nonsense").code(404))
```

rule-types/github/branch_protection_allow_force_pushes_test.star
```diff
@@ -0,0 +1,46 @@
+def test_force_push_not_allowed():
+    res = eval(
+        rule="branch_protection_allow_force_pushes",
+        entity={"owner": "mindersec", "name": "minder", "type": "repository", "default_branch": "main"},
+        profile={},
+        mock_http={
+            "/repos/mindersec/minder/branches/main/protection": body("{\"allow_force_pushes\": {\"enabled\": false}}")
```
evankanderson: Given how often `body` is JSON content, I wonder if we want `body` to be able to accept an object and JSON-encode it.

rule-types/github/branch_protection_allow_force_pushes_test.star
```diff
@@ -0,0 +1,46 @@
+def test_force_push_not_allowed():
+    res = eval(
+        rule="branch_protection_allow_force_pushes",
+        entity={"owner": "mindersec", "name": "minder", "type": "repository", "default_branch": "main"},
+        profile={},
```
evankanderson: Do we need to fill in the `profile` argument at all?

rule-types/github/branch_protection_allow_force_pushes_test.star
```diff
@@ -0,0 +1,46 @@
...
+        mock_http={
+            "/repos/mindersec/minder/branches/main/protection": body("{\"allow_force_pushes\": {\"enabled\": true}}")
+        }
+    )
+    assert.true(res["status"] in ("fail", "error"))
+    assert.true(res["message"] != "")
```
evankanderson: I've been seeing this for a bit -- I think testing the error message would be fine, but this "is not empty" check seems not very useful to me.

rule-types/github/branch_protection_allow_force_pushes_test.star
```diff
@@ -0,0 +1,46 @@
...
+
+def test_not_found():
+    res = eval(
+        rule="branch_protection_allow_force_pushes",
+        entity={"owner": "mindersec", "name": "minder", "type": "repository", "default_branch": "main"},
+        profile={},
+        mock_http={
+            "/repos/mindersec/minder/branches/main/protection": body("{\"woot\": \"woot\"}").code(404)
```
evankanderson: Do we need a body (or a non-empty body) to be able to send a 404?

rule-types/github/branch_protection_require_status_checks_test.star
```diff
@@ -0,0 +1,57 @@
+def test_status_checks_required():
+    res = eval(
+        rule="branch_protection_require_status_checks",
+        entity={"owner": "mindersec", "name": "minder", "type": "repository", "default_branch": "main"},
+        profile={},
+        mock_http={
+            "/repos/mindersec/minder/branches//protection": body("{\"required_status_checks\": {\"url\": \"https://api.github.com/repos/test/test/branches/main/protection/required_status_checks\", \"strict\": false, \"contexts\": [], \"checks\": []}}")
```
evankanderson: The body content here is complex enough that it really feels like it should stretch to multiple lines.

evankanderson: Is the double-slash with no branch name expected / correct?

rule-types/github/branch_protection_require_status_checks_test.star
```diff
@@ -0,0 +1,57 @@
...
+    )
+    assert.eq(res["status"], "pass")
+
+def test_status_checks_not_required():
+    res = eval(
+        rule="branch_protection_require_status_checks",
+        entity={"owner": "mindersec", "name": "minder", "type": "repository", "default_branch": "main"},
+        profile={},
+        mock_http={
+            "/repos/mindersec/minder/branches//protection": body("{\"required_status_checks\": null}")
```
evankanderson: We may need some special affordances for serializing JSON if we want to be able to pass an object to `body` here.

rule-types/github/golangci-lint_github_action_test.star
```diff
@@ -0,0 +1,22 @@
...
+def test_should_have_golangci_lint_gh_action_configured():
+    res = eval(
+        rule="golangci-lint_github_action",
+        entity={"type": "repository", "default_branch": "main"},
+        profile={},
+        mock_fs={
+            ".github/workflows/positive_mock.yml": read_file("golangci-lint_github_action.testdata/repo_with_golangci-lint_gh_action/.github/workflows/positive_mock.yml")
+        }
+    )
+    assert.eq(res["status"], "pass")
```
evankanderson, line 21: Since we no longer need to match paths, what do you think about simplifying these paths (or maybe follow up by migrating these to a single `.txtar` in a `testdata` directory)?

rule-types/github/repo_issues_enabled_test.star
```diff
@@ -0,0 +1,46 @@
...
...\"has_pages\": false,\n    \"has_discussions\": false,\n    \"forks_count\": 5,\n    \"mirror_url\": null,\n    \"archived\": false,\n    \"disabled\": false,\n    \"open_issues_count\": 12,\n    \"license\": {\n      \"key\": \"apache-2.0\",\n      \"name\": \"Apache License 2.0\",\n      \"spdx_id\": \"Apache-2.0\",\n      \"url\": \"https://api.github.com/licenses/apache-2.0\",\n      \"node_id\": \"MDc6TGljZW5zZTI=\"\n    },\n    \"allow_forking\": true,\n    \"is_template\": false,\n    \"web_commit_signoff_required\": false,\n    \"topics\": [],\n    \"visibility\": \"public\",\n    \"forks\": 5,\n    \"open_issues\": 12,\n    \"watchers\": 2,\n    \"default_branch\": \"main\"\n  },\n  \"security_and_analysis\": {\n    \"secret_scanning\": {\n      \"status\": \"disabled\"\n    },\n    \"secret_scanning_push_protection\": {\n      \"status\": \"disabled\"\n    },\n    \"dependabot_security_updates\": {\n      \"status\": \"disabled\"\n    },\n    \"secret_scanning_non_provider_patterns\": {\n      \"status\": \"disabled\"\n    },\n    \"secret_scanning_validity_checks\": {\n      \"status\": \"disabled\"\n    }\n  },\n  \"network_count\": 5,\n  \"subscribers_count\": 1\n}\n")
```
evankanderson: Most of the contents of the JSON response aren't needed by the test.  Can we either:

1. Simplify the JSON to the elements needed **OR**
2. Use `read_file` to load the existing verbatim JSON contents rather than having them in the file?

rule-types/github/trufflehog_github_action_test.star
```diff
@@ -0,0 +1,22 @@
...
+        rule="trufflehog_github_action",
+        entity={"type": "repository", "default_branch": "main"},
+        profile={},
+        mock_fs={
+            ".github/workflows/trufflehog.yaml": read_file("trufflehog_github_action.testdata/github_action_with_trufflehog/.github/workflows/trufflehog.yaml")
+        }
+    )
+    assert.eq(res["status"], "pass")
+
+def test_should_not_have_renovate_enabled():
```
evankanderson: I realize we're directly transcribing test names, but this should probably be `test_should_not_have_trufflehog_enabled`.

evankanderson (commented):
This is exciting!  A bunch of comments seeing actual usage (and we may want a v0.2.1 with some additional functionality), but this looks great (particularly the -2570 lines!)

rule-types/common/enforce_file_test.star
evankanderson, line 1: What do you think of the following?  Is it too terse?

```python
def PASS(res):
  assert.eq(res["status"], "pass")

def FAIL(res):
  assert.true(res["status"] in "fail", "error")  # TODO: should we match both?  Error and failure are different
  assert.true(res["message"] != "")  # TODO: should we be checking the message contents?  If not, is this check helpful?

def enforce_file(content, fs):
  return eval(
    rule="enforce_file",
    entity={"type": "repository", "default_branch": "main"},
    profile={”file”: “README”, “content”: content},
    mock_fs=fs
  )

file_present = {
  “README”: read_file("enforce_file.testdata/file_present/README")
}

def test_file_should_be_present():
  PASS(enforce_file(“”, file_present)

def test_file_is_missing():
  res = enforce_file(“”, {})
  FAIL(res)

def test_file_present_and_matches_content():
  PASS(enforce_file(“...”, file_present)

def test_file_present_but_has_different_content():
  res = enforce_file(“Different”, {})
  FAIL(res)

def test_file_present_but_more_content():
  res = enforce_file(“Test”, {})
  FAIL(res)

def test_file_present_but_less_content():
  res = enforce_file(“Test content with more”, {})
  FAIL(res)
```

rule-types/common/python_lock_file_exists_test.star
evankanderson: (It's fine to say something like "I want to leave that diff for a later PR to make this easier to review" on comments like this.)

rule-types/common/require_pre_commit_to_be_configured_test.star
```diff
@@ -0,0 +1,32 @@
+def test_should_have_at_least_one_pre_commit_hook_configured():
+    res = eval(
+        rule="require_pre_commit_to_be_configured",
+        entity={"type": "repository", "default_branch": "main"},
+        profile={},
```
evankanderson: Do we need the `profile` argument at all here (and previous)?

rule-types/github/branch_protection_allow_deletions_test.star
```diff
@@ -0,0 +1,61 @@
+ENTITY = {"owner": "mindersec", "name": "minder", "type": "repository", "default_branch": "main"}
+URL = "/repos/mindersec/minder/branches/main/protection"
+
+def PASS(res):
+    assert.eq(res["status"], "pass")
+
+def FAIL(res):
+    assert.true(res["status"] in ("fail", "error"))
+    assert.true(res["message"] != "")
```
evankanderson: Do we want to try to extract these PASS/FAIL definitions out into a shared module that's imported?  (This can be another "yes, but not in this PR" bit of feedback if that's what you want -- finding the cut points on a series of changes can be a bit of an art form.)

rule-types/github/branch_protection_allow_deletions_test.star
```diff
@@ -0,0 +1,61 @@
...
+    res = eval(
+        rule="branch_protection_allow_deletions",
+        entity=ENTITY,
+        profile={},
+        mock_http={
+            URL: body("""{
+              "allow_deletions": {
+                "enabled": false
+              }
+            }""")
```
evankanderson: Triple-quotes (or single-quotes) feels better than escaped double-quotes, thanks!

rule-types/github/repo_issues_enabled_test.star
```diff
@@ -0,0 +1,57 @@
...
+
+def test_issues_are_enabled():
+    res = eval(
+        rule="repo_issues_enabled",
+        entity=ENTITY,
+        mock_http={
+            URL: body("""{
+              "has_issues": true
+            }""")
+        }
```
evankanderson, line 17: Why not use `read_file` to load these bodies?

rule-types/github/repo_visibility_test.star
```diff
@@ -0,0 +1,57 @@
...
...\"has_pages\": false,\n    \"has_discussions\": false,\n    \"forks_count\": 5,\n    \"mirror_url\": null,\n    \"archived\": false,\n    \"disabled\": false,\n    \"open_issues_count\": 12,\n    \"license\": {\n      \"key\": \"apache-2.0\",\n      \"name\": \"Apache License 2.0\",\n      \"spdx_id\": \"Apache-2.0\",\n      \"url\": \"https://api.github.com/licenses/apache-2.0\",\n      \"node_id\": \"MDc6TGljZW5zZTI=\"\n    },\n    \"allow_forking\": true,\n    \"is_template\": false,\n    \"web_commit_signoff_required\": false,\n    \"topics\": [],\n    \"visibility\": \"public\",\n    \"forks\": 5,\n    \"open_issues\": 12,\n    \"watchers\": 2,\n    \"default_branch\": \"main\"\n  },\n  \"security_and_analysis\": {\n    \"secret_scanning\": {\n      \"status\": \"disabled\"\n    },\n    \"secret_scanning_push_protection\": {\n      \"status\": \"disabled\"\n    },\n    \"dependabot_security_updates\": {\n      \"status\": \"disabled\"\n    },\n    \"secret_scanning_non_provider_patterns\": {\n      \"status\": \"disabled\"\n    },\n    \"secret_scanning_validity_checks\": {\n      \"status\": \"disabled\"\n    }\n  },\n  \"network_count\": 5,\n  \"subscribers_count\": 1\n}\n")
```
evankanderson: Ditto on large payloads with JSON.  I'd sort of try the following decision tree:

1. If the JSON is greater than 8 lines pretty-printed, read from external file.  (This can include dumps from the GitHub API)
2. If the JSON is more than 3 lines pretty-printed, triple-quotes are probably a good mechanism to help delimit start/end of the body.
3. Use single-quotes to avoid needing to escape double quotes, and avoid encoding newlines in the JSON, to minimize the amount of escaped characters.

evankanderson (commented):
A few more comments, but this is already feeling better to me in terms of understanding and reasoning about the contents without needing to go reference the underlying implementation code to figure out what's expected.

krrish175-byte:
@dakshhhhh16 @evankanderson PTAL

.github/workflows/test.yml
```diff
@@ -14,26 +14,8 @@ jobs:
       - name: Checkout
         uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v4
 
-      - name: Set up Go
-        uses: actions/setup-go@924ae3a1cded613372ab5595356fb5720e22ba16 # v5
+      - name: Run tests
+        uses: mindersec/minder-action/test@main
```
evankanderson: Let's get a working release and use that, but feel free to reference in-progress branches while getting tests to pass.

.github/workflows/test.yml
```diff
@@ -14,26 +14,8 @@ jobs:
...
-          token: ${{ secrets.GITHUB_TOKEN }}
-
-      - name: Run go mod download
-        run: go mod download
-
-      - name: build
-        run: |-
-          task test
+          minder_version: v0.2.2
+          rules_directory: rule-types
```
evankanderson, line 22: Do we want to run a formatter after this completes?

Taskfile.yml
```diff
@@ -19,13 +19,15 @@ tasks:
...
       
       Note you must also provide a test data directory that contains the test data files.
       The test data directory must be named as the rule type file, but with a `.testdata`
       suffix. For example, if the rule type is called `my_rule_type.yaml` then the test data
       directory must be called `my_rule_type.testdata`.
+      
+      You must have `mindev` installed and available in your PATH to run tests.
     cmds:
-      - go test -json -v ./... | gotestfmt -hide "all"
+      - mindev test rule-types/**/*_test.star
```
evankanderson: Can't this just be `mindev test rule-types`?  Is the other (Ruby) glob syntax actually working?

go.mod
evankanderson, line 1: Do we still need a `go.mod`/`go.sum`?

rule-types/common/python_lock_file_exists_test.star
evankanderson: I'm assuming this is a "defer to the future"?

rule-types/github/repo_issues_enabled_test.star
evankanderson: Still wondering about this; you could use a helper to encode a lot of the path wordiness.

evankanderson (commented):
A few unresolved comments from earlier; I like how a bunch of the starlark tests can be shorter than the corresponding yaml while more clearly highlighting the important differences between the test cases.

.github/workflows/test.yml
evankanderson:
```suggestion
        uses: mindersec/minder-action/test@v1.1.0
```

rule-types/github/repo_visibility.testdata/private.json
```diff
@@ -226,111 +226,6 @@
     "watchers": 2,
     "default_branch": "main"
   },
-  "source": {
```
evankanderson, line 229: Why were these removed?

rule-types/github/repo_visibility_test.star
```diff
@@ -0,0 +1,66 @@
...
...  \"has_pages\": false,\n    \"has_discussions\": false,\n    \"forks_count\": 5,\n    \"mirror_url\": null,\n    \"archived\": false,\n    \"disabled\": false,\n    \"open_issues_count\": 12,\n    \"license\": {\n      \"key\": \"apache-2.0\",\n      \"name\": \"Apache License 2.0\",\n      \"spdx_id\": \"Apache-2.0\",\n      \"url\": \"https://api.github.com/licenses/apache-2.0\",\n      \"node_id\": \"MDc6TGljZW5zZTI=\"\n    },\n    \"allow_forking\": true,\n    \"is_template\": false,\n    \"web_commit_signoff_required\": false,\n    \"topics\": [],\n    \"visibility\": \"public\",\n    \"forks\": 5,\n    \"open_issues\": 12,\n    \"watchers\": 2,\n    \"default_branch\": \"main\"\n  },\n  \"security_and_analysis\": {\n    \"secret_scanning\": {\n      \"status\": \"enabled\"\n    },\n    \"secret_scanning_push_protection\": {\n      \"status\": \"enabled\"\n    },\n    \"dependabot_security_updates\": {\n      \"status\": \"disabled\"\n    },\n    \"secret_scanning_non_provider_patterns\": {\n      \"status\": \"disabled\"\n    },\n    \"secret_scanning_validity_checks\": {\n      \"status\": \"disabled\"\n    }\n  },\n  \"network_count\": 5,\n  \"subscribers_count\": 1\n}\n")
```
evankanderson: Rather than using `body("lots-of-json")`, let's use `body(read_file(...))` and put the JSON in its own file (particularly since these files are already checked in).

In the future, we could add a comment next to the file saying something like "generated by gh api /repo/... > filename" to make it easier to re-create if needed.

evankanderson (dismissed):
Okay, I think the CI should start passing now that we have all the releases lined up.

rule-types/github/secret_scanning_test.star
```diff
@@ -0,0 +1,68 @@
...
...  \"has_pages\": false,\n    \"has_discussions\": false,\n    \"forks_count\": 5,\n    \"mirror_url\": null,\n    \"archived\": false,\n    \"disabled\": false,\n    \"open_issues_count\": 12,\n    \"license\": {\n      \"key\": \"apache-2.0\",\n      \"name\": \"Apache License 2.0\",\n      \"spdx_id\": \"Apache-2.0\",\n      \"url\": \"https://api.github.com/licenses/apache-2.0\",\n      \"node_id\": \"MDc6TGljZW5zZTI=\"\n    },\n    \"allow_forking\": true,\n    \"is_template\": false,\n    \"web_commit_signoff_required\": false,\n    \"topics\": [],\n    \"visibility\": \"public\",\n    \"forks\": 5,\n    \"open_issues\": 12,\n    \"watchers\": 2,\n    \"default_branch\": \"main\"\n  },\n  \"security_and_analysis\": {\n    \"secret_scanning\": {\n      \"status\": \"enabled\"\n    },\n    \"secret_scanning_push_protection\": {\n      \"status\": \"enabled\"\n    },\n    \"dependabot_security_updates\": {\n      \"status\": \"disabled\"\n    },\n    \"secret_scanning_non_provider_patterns\": {\n      \"status\": \"disabled\"\n    },\n    \"secret_scanning_validity_checks\": {\n      \"status\": \"disabled\"\n    }\n  },\n  \"network_count\": 5,\n  \"subscribers_count\": 1\n}\n")
```
evankanderson: Another case for `body(read_file(".../enabled.json"))`

.github/workflows/test.yml
evankanderson: https://github.com/mindersec/minder-action suggests something like the following:


```suggestion
          rules_directory: rule-types
        id: run-tests

      - name: Publish Test Report
        uses: mikepenz/action-junit-report@v4
        if: success() || failure() # always run even if tests fail
        with:
          report_paths: ${{ steps.run-tests.outputs.junit-filename }}```

**Review by evankanderson — COMMENTED** [review]
This is looking close once we get a new Minder release cut (hopefully today after merging https://github.com/mindersec/minder/pull/6639 )

File: .github/workflows/test.yml [hunk 28]
```diff
@@ -14,26 +14,15 @@ jobs:
...
 
-      - name: Set up Go
-        uses: actions/setup-go@924ae3a1cded613372ab5595356fb5720e22ba16 # v5
+      - name: Run tests
+        uses: mindersec/minder-action/test@v1.1.0
+        id: run-tests
         with:
-          go-version-file: 'go.mod'
-          check-latest: true
+          minder_version: v0.2.2
```
evankanderson:
```suggestion
          minder_version: v0.3.0
```

evankanderson (commented):
I think you need to update the test workflow's minder version, and there are a couple other failing workflows that need some fix.  (Even if it is a "TODO" to re-enable them.)

evankanderson:
Looking into the other failures, now that `mindev` is part of the release bundle, you'll need to change the following (for example) to use `minder-client-installer` and then just run `mindev ruletype lint`:

```yaml
      - name: Set up Go
        uses: actions/setup-go@924ae3a1cded613372ab5595356fb5720e22ba16 # v5
        with:
          go-version-file: 'go.mod'
          check-latest: true

      - name: Lint Rule Types
        run: go run github.com/mindersec/minder/cmd/dev@latest ruletype lint -r rule-types/github --skip-rego
```

.github/workflows/lint.yaml
```diff
@@ -14,15 +14,20 @@ jobs:
       - name: Checkout
         uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v4
 
+      - name: Install cosign
+        uses: sigstore/cosign-installer@v3.6.0
```
evankanderson: This needs to be:


```suggestion
        uses: sigstore/cosign-installer@v4.1.2
```

(Or some other v4 version to install the v3 of cosign)

.github/workflows/update.yaml
```diff
@@ -22,11 +22,11 @@ jobs:
         ref: main
         path: before_files
 
-    - name: Set up Go
-      uses: actions/setup-go@924ae3a1cded613372ab5595356fb5720e22ba16 # v5
-      with:
-        go-version-file: 'go.mod'
-        check-latest: true
+    - name: Install cosign
+      uses: sigstore/cosign-installer@v3.6.0
```
evankanderson: Ditto here:


```suggestion
        uses: sigstore/cosign-installer@v4.1.2
```

.github/workflows/lint.yaml
```diff
@@ -14,15 +14,20 @@ jobs:
...
+        uses: mindersec/minder-client-installer@430dc4a6f07d6645288b5812fe9b780117963cf2 # v1.1.2
+
+      - name: Lint Rule Types
+        run: mindev ruletype lint -r rule-types/github --skip-rego
+
       - name: Set up Go
         uses: actions/setup-go@924ae3a1cded613372ab5595356fb5720e22ba16 # v5
         with:
-          go-version-file: 'go.mod'
+          go-version: '1.22'
```
evankanderson: OPA requires Go 1.25, but there's also an `open-policy-agent/setup-opa@v2` action that will install OPA without needing to compile it.

```suggestion
          go-version: '1.26'
```

evankanderson:
It also looks like the "pass" versions of the following tests are unexpected failing:

```
FAIL: branch_protection_allow_deletions_test.star/test_force_push_not_allowed
FAIL: branch_protection_allow_force_pushes_test.star/test_force_push_not_allowed
FAIL: branch_protection_require_status_checks_test.star/test_status_checks_required
FAIL: branch_protection_require_status_checks_test.star/test_status_checks_required_with_strict_mode
```

All of these seem to have the following pattern on ingest:

```
  ingest:
    type: rest
    rest:
      endpoint: '{{ $branch_param := index .Params "branch" }}/repos/{{.Entity.Owner}}/{{.Entity.Name}}/branches/{{if ne $branch_param "" }}{{ $branch_param }}{{ else }}{{ .Entity.DefaultBranch }}{{ end }}/protection'
 ```

Which seems to be a complicated way of trying to fetch the `params.branch` field and use either it or the `default_branch` field on the entity.

File: .github/workflows/test.yml [hunk 32]
```diff
@@ -14,26 +14,15 @@ jobs:
...
-      - name: Set up Go
-        uses: actions/setup-go@924ae3a1cded613372ab5595356fb5720e22ba16 # v5
+      - name: Run tests
+        uses: mindersec/minder-action/test@v1.1.0
+        id: run-tests
         with:
-          go-version-file: 'go.mod'
-          check-latest: true
+          minder_version: v0.3.0
+          rules_directory: rule-types
```
evankanderson: I renamed this to `release` (and `rules-directory` in https://github.com/mindersec/minder-action/pull/4):


```suggestion
          release: v0.3.0
          rules-directory: rule-types
```

evankanderson (commented):
I think one change (and maybe a v0.3.1 release of Minder), and this should be good to go.

.github/workflows/test.yml
```diff
@@ -14,26 +14,15 @@ jobs:
...
 
-      - name: Set up Go
-        uses: actions/setup-go@924ae3a1cded613372ab5595356fb5720e22ba16 # v5
+      - name: Run tests
+        uses: mindersec/minder-action/test@v1.1.0
+        id: run-tests
         with:
-          go-version-file: 'go.mod'
-          check-latest: true
+          release: v0.3.0
```
evankanderson:
```suggestion
          release: v0.3.1
```

I think you need https://github.com/mindersec/minder/pull/6688 for the branch protection tests to pass without passing in a params map (which I think is probably the right way to fix it).

evankanderson (commented):
One more fix to get the `test` run to pass cleanly, I think.

evankanderson (approved):
Excellent!  I'm looking forward to not having any of the CI in this repository need or depend on a `go.mod` (removing dependabot requirements and speeding up the CI).

Thanks!

PR #419, opened 2026-07-18
feat: add GitLab branch protection and change history OSPS rules

intelligent-ears:
## Summary

Adds 2 more GitLab OSPS level-1 rule types, continuing the work from #412.

## Rules added

| Rule ID | Description | GitLab API | Endpoint |
|---------|-------------|------------|----------|
| osps-ac-03-01 | Prevent overwriting git history | `allow_force_push == false` | `GET /projects/:id/protected_branches/:branch` |
| osps-qa-01-02 | Maintain publicly readable change history | `is_private == false` + `allow_force_push == false` | `GET /projects/:id/protected_branches/:branch` |

Both rules validated end-to-end against live GitLab projects using a local Minder stack.

## Notes

- GitLab's Protected Branches API returns `allow_force_push` as a direct boolean — simpler than GitHub's equivalent which also requires checking rulesets via a data source
- `osps-ac-03-02` (prevent branch deletion) is **not included** — GitLab doesn't have a direct `allow_deletions` boolean equivalent. Branch deletion protection in GitLab is controlled through `unprotect_access_levels` semantics which differ meaningfully from GitHub's model. Flagging for discussion before implementing.

## Part of

Continues GitLab OSPS level-1 rule porting — part of mindersec/minder#6435

### DIFF
--- a/security-baseline/rule-types/gitlab/osps-ac-03-01.yaml
+++ b/security-baseline/rule-types/gitlab/osps-ac-03-01.yaml
@@ -1,0 +1,46 @@
+version: v1
+release_phase: alpha
+type: rule-type
+name: osps-ac-03-01
+display_name: Prevent overwriting git history
+short_failure_message: Force pushes are allowed
+severity:
+  value: info
+description: Disallow force pushes to the default branch
+guidance: |
+  Ensure that force pushes are disabled for the default branch
+  protection rule.
+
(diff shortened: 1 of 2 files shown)


evankanderson (approved):
Not for this set, but you should look at #418 for the future testing tools that should help verify these types of rules.

PR #430, opened 2026-08-11
test: add starlark tests for github rule types

krrish175-byte:
Adds comprehensive Starlark unit tests for several GitHub rule types, including `secret_push_protection`, `automatic_branch_deletion`, and `branch_protection_enforce_admins`.


evankanderson:
You'll need to merge main now that #418 has been merged

rule-types/github/branch_protection_enforce_admins_test.star
```diff
@@ -0,0 +1,45 @@
...
+        profile={"enforce_admins": True},
+        mock_http=build_mock_http(True)
+    )
+    assert.eq(res["status"], "pass")
+
+def test_branch_protection_enforce_admins_disabled():
+    res = eval(
+        rule=RULE,
+        entity=ENTITY,
+        profile={"enforce_admins": True},
```
evankanderson, line 35: This is an older rule, where we ended up using more parameters than we do now (e.g. why `enforce_admins: false` as a policy?).  Regardless, it would be good to test `{"branch": "other"}` and `{"enforce_admins": False}` cases.

rule-types/github/secret_push_protection_test.star
```diff
@@ -0,0 +1,49 @@
...
+    "name": "minder",
+    "type": "repository",
+    "default_branch": "main"
+}
+
+RULE = "secret_push_protection"
+
+def build_mock_http(is_private, status):
+    return {
+        "/repos/mindersec/minder": body('{"private": ' + ("true" if is_private else "false") + ', "security_and_analysis": {"secret_scanning_push_protection": {"status": "' + status + '"}}}')
```
evankanderson: This line feels too long.  I'm not also not thrilled with either option I see for the string replacement:

* String addition / join -- makes it hard to see / track the curly braces
* [`string.format`](https://bazel.build/rules/lib/core/string#format) -- requires doubling curly braces to quote them

Thoughts?  (Since JSON is a pretty common case, it would be nice for it to be fairly smooth.

evankanderson (commented):
This looks reasonable, though I'm wondering whether it makes sense to combine a bunch of the "repo status checks" and "branch status checks" tests into a single file that tests multiple rules.  (Still thinking it through... there are pros and cons.)

What do you think?

evankanderson (approved):
Thanks, these look great!

evankanderson (changes requested):
Oops, one test is failing: 

```
Error(s) running tests:
error running file rule-types/github/branch_protection_enforce_admins_test.star: loading rule-types/github/branch_protection_enforce_admins_test.star: rule-types/github/branch_protection_enforce_admins_test.star:17:73: undefined: json
error running file rule-types/github/secret_push_protection_test.star: loading rule-types/github/secret_push_protection_test.star: rule-types/github/secret_push_protection_test.star:20:41: undefined: json
```

evankanderson (approved):

PR #431, opened 2026-08-16
test: add starlark tests for github branch protection and secret rules

krrish175-byte:
## Description

This PR adds automated Starlark test coverage for several GitHub rule types in the `minder-rules-and-profiles` repository to ensure their JQ and Rego evaluation logic works as expected across various configurations. 

Specifically, this adds tests for the following rules:
- `automatic_branch_deletion`
- `branch_protection_allow_fork_syncing`
- `branch_protection_enabled`
- `branch_protection_enforce_admins`
- `branch_protection_lock_branch`
- `branch_protection_require_conversation_resolution`
- `branch_protection_require_linear_history`
- `branch_protection_require_pull_request_approving_review_count`
- `branch_protection_require_pull_request_code_owners_review`
- `branch_protection_require_pull_request_dismiss_stale_reviews`
- `branch_protection_require_pull_request_last_push_approval`
- `branch_protection_require_pull_requests`
- `branch_protection_require_signatures`
- `secret_push_protection`

## Changes
- Created 14 new `*_test.star` files containing unit tests for the rule logic.
- Included comprehensive test scenarios for each rule, including:
  - Enabled / Disabled states returned from the GitHub API.
  - Unprotected branches (HTTP 404 responses).
  - Dynamic rule profile configurations (e.g., verifying a rule passes if the profile sets the requirement to `false` and the API returns `false`).
  - Checking custom branch names via parameter passing.

## Verification
- Ran `mindev test ./rule-types/github` locally and verified that all newly added test suites pass successfully.


rule-types/github/branch_protection_enforce_admins_test.star
evankanderson: What is the actual GitHub message here?  I seem to get something like:

```
$ gh api /repos/evankanderson/minder/branches/admin-delete/protection
{
  "message": "Branch not protected",
  "documentation_url": "https://docs.github.com/rest/branches/branch-protection#get-branch-protection",
  "status": "404"
}
gh: Branch not protected (HTTP 404)
```

evankanderson (dismissed):
Thanks for adding these!

Any notion as to what fraction of the rules we have coverage for now?

Anything you're wanting to change about `mindev test` before writing more rules?

krrish175-byte:
Thanks for catching that! I have updated the `mock_http_404` payloads in all the branch protection tests to correctly mock the actual GitHub API response:
`{"message": "Branch not protected", "documentation_url": "https://docs.github.com/rest/branches/branch-protection#get-branch-protection", "status": "404"}`

As for your other questions:

1. **Test Coverage:** With this PR, we have tests for 29 out of the 69 total rules, which puts us at roughly **42% coverage**. 

2. **Thoughts on `mindev test` enhancements:** 
   - **Starlark `load()` support:** It would be super helpful to be able to load external `.star` files so we can share common test fixtures. Right now, things like the `ENTITY` struct and the `build_mock_http` functions are being copy-pasted across multiple test files. Being able to extract those into a shared `github_mocks.star` file would make writing and maintaining tests a lot cleaner.
   - **Boilerplate generation:** An option to auto-generate the test boilerplate (entity, basic mock, evaluation struct) for a given rule would save a lot of time.
   - **Better error visibility:** It would be nice to have slightly more granular error reporting when things like JQ paths fail during evaluation.

rule-types/github/branch_protection_test.star
```diff
@@ -0,0 +1,763 @@
...
+    return {
+        "/repos/mindersec/minder/branches/%s/protection" % branch: body(payload).code(404)
+    }
+
+
+def build_mock_http_branch_protection_allow_fork_syncing(enabled, branch="main"):
+    payload = '{"allow_fork_syncing": {"enabled": %s}}' % ("true" if enabled else "false")
+    return {
+        "/repos/mindersec/minder/branches/%s/protection" % branch: body(payload)
+    }
```
evankanderson, line 54: kwargs might make this simpler (along with an object -> string converter):

```python
def to_json(data):
    """Convert a starlark (Python-style) object to JSON.  Returns a string."""
    if type(data) == "None":
        return "null"
    if type(data) == "bool":
        return "true" if data else "false"
    if type(data) == "string":
        return '"%s"' % data
    if type(data) == "int" or type(data) == "float":
        return "%d" % data
    # We can't call functions recursively in starlark
    # Generate an error
    assert.true(False)

def json_node(data):
    """Generate JSON from a node which is either a dict or a string.  Returns a string.
   This is a workaround for no recursive calls in starlark."""
    if type(data) == "string":
        return '"%s"' % data
    if type(data) == "dict":
        return "{%s}" % ",".join(['%s : %s' % (to_json(k), to_json(v)) for (k, v) in data.items()])
    assert.true(False)

def build_branch_response(branch="main", **updates):
    base = {
        "url": "https://api.github.com/repos/mindersec/minder/branches/%s/protection" % branch,
        "allow_fork_syncing": {"enabled": True},
        "enforce_admins": {"enabled": True},
        "lock_branch": {"enabled": True},
        "required_conversation_resolution": {"enabled": True},
        "required_linear_history": {"enabled": True},
        "required_signatures": {"enabled": True},
        "required_pull_request_reviews": {
            "required_approving_review_count": 0,
            "require_code_owner_reviews": True,
            "dismiss_stale_reviews": True,
            "require_last_push_approval": True,
            "url": "https://api.github.com/repos/mindersec/minder/branches/%s/protection/required_pull_request_reviews" % branch,
        },
    }
    if updates:
        base.update(updates)
    payload = "{%s}" % ", ".join(['"%s" : %s' % (k, json_node(v)) for (k, v) in base.items()])
    return {
        "/repos/mindersec/minder/branches/%s/protection" % branch: body(payload)
    }
```

And then:

```python
def test_branch_protection_allow_fork_syncing_disabled():
    res = eval(
        rule="branch_protection_allow_fork_syncing",
        entity=ENTITY,
        profile={"allow_fork_syncing": True},
        mock_http=build_branch_response(allow_fork_syncing={"enabled": False})
    )
    assert.eq(res["status"], "fail")
```

etc, rather than needing to define different `build_mock_http_branch_protection_...` methods for each function.

There's also a simpler kwargs version that only works for the `{"enabled": True}` response keys if you don't want as much complexity.

rule-types/github/branch_protection_test.star
```diff
@@ -0,0 +1,763 @@
...
+
+def test_branch_protection_allow_fork_syncing_enabled():
+    res = eval(
+        rule="branch_protection_allow_fork_syncing",
+        entity=ENTITY,
+        profile={"allow_fork_syncing": True},
+        mock_http=build_mock_http_branch_protection_allow_fork_syncing(True)
+    )
```
evankanderson: If the eval / assert status boilerplate seems too much, you can do something like the following:

```python
def test_fork_sync_cases():
    cases = {
        "enabled": {},
        "disabled": {"got": False, "fail": True},
        "unprotected": {"body": build_mock_http_404, "fail": True},
        "false_enabled": {"want": False, "fail": True},
        "false_disabled": {"want": False, "got": False},
        "custom_branch": {"branch": "other"}}
    
    for case, overrides in cases.items():
        values = {"want": True, "got": True}
        values.update(overrides)
        params = {"branch": values["branch"]} if "branch" in values else {}
        if "body" in values:
            http = values["body"]()
        else:
            http = build_branch_response(allow_fork_syncing={"enabled": values["got"]}, branch=values.get("branch", "main"))
        res = eval(
            rule="branch_protection_allow_fork_syncing",
            entity=ENTITY,
            params=params,
            profile={"allow_fork_syncing": values["want"]},
            mock_http=http,
        )
        assert.eq(res["status"], "fail" if "fail" in values else "pass")
```

But I'm not sure that level of DRY helps.  (To your point about needing to repeat `ENTITY` and `rule`)

evankanderson: Actually, you might be able to use this pattern to "stamp out" tests for the `{"enabled": True/False}` settings, which would only leave you with `required_pull_request_reviews` and `branch_protection_enabled` tests to run.

evankanderson (commented):
I'm okay with merging this as-is if you prefer it, but a few ideas on making this less verbose. (I cloned your fork to play with the various options -- not being able to do recursion in Starlark makes it hard to build a nice general-purpose `to_json` method.)

rule-types/github/branch_protection_test.star
```diff
@@ -0,0 +1,334 @@
...
+            
+        res = eval(
+            rule=rule_name,
+            entity=ENTITY,
+            params=params,
+            profile=profile,
+            mock_http=http,
+        )
+        expected_status = "fail" if "fail" in values else "pass"
+        # Include case name in the assert to make it easy to debug
```
evankanderson, line 41: ❤️

rule-types/github/branch_protection_test.star
```diff
@@ -0,0 +1,334 @@
...
+        assert.eq("%s:%s" % (case, res["status"]), "%s:%s" % (case, expected_status))
+def build_mock_http_404(branch="main"):
```
evankanderson, line 43: Put some space between the definitions
```suggestion


def build_mock_http_404(branch="main"):
```

rule-types/github/branch_protection_test.star
```diff
@@ -0,0 +1,334 @@
...
+    return {
+        "/repos/mindersec/minder/branches/%s/protection" % branch: body(payload).code(404)
+    }
+
+
+def build_mock_http_branch_protection_allow_fork_syncing(enabled, branch="main"):
+    payload = '{"allow_fork_syncing": {"enabled": %s}}' % ("true" if enabled else "false")
+    return {
+        "/repos/mindersec/minder/branches/%s/protection" % branch: body(payload)
+    }
```
evankanderson, line 54: It's weird to have this function up here, but the other similar functions after the `branch_protection_enabled` case.  Maybe move this down to line 106-ish?

rule-types/github/branch_protection_test.star
```diff
@@ -0,0 +1,334 @@
...
+        mock_http=build_mock_http_branch_protection_require_pull_requests(True, branch="other")
+    )
+    assert.eq(res["status"], "pass")
+
+
+def build_mock_http_branch_protection_require_signatures(enabled, branch="main"):
+    payload = '{"required_signatures": {"enabled": %s}}' % ("true" if enabled else "false")
+    return {
+        "/repos/mindersec/minder/branches/%s/protection" % branch: body(payload)
+    }
```
evankanderson, line 269: This is the same pattern as require_linear_history, right?  Move it up by that on line 128?

rule-types/github/branch_protection_test.star
```diff
@@ -0,0 +1,334 @@
...
+
+def test_branch_protection_require_signatures_custom_branch():
+    res = eval(
+        rule="branch_protection_require_signatures",
+        entity=ENTITY,
+        params={"branch": "other"},
+        profile={},
+        mock_http=build_mock_http_branch_protection_require_signatures(True, branch="other")
+    )
+    assert.eq(res["status"], "pass")
```
evankanderson, line 309: These could use `run_boolean_rule_tests`, right?

rule-types/github/branch_protection_test.star
```diff
@@ -0,0 +1,334 @@
...
+
+
+def test_branch_protection_allow_fork_syncing():
+    run_boolean_rule_tests("branch_protection_allow_fork_syncing", "allow_fork_syncing", build_mock_http_branch_protection_allow_fork_syncing)
```
evankanderson, line 313: Since you use `run_boolean_rule_tests` down here, I'd be inclined to move that definition along with these `build_mock_...` functions down here.

I'd also include a comment explaining why we've got this machinery for building tests here but not elsewhere, something along the lines of "There are 7 different rules which each manage a specific part of branch protection rules.  GitHub has structured these as 'key: {enabled: true}' JSON objects in the branch protection resource, so we aim to test the same cases for each of these different keys".

In general, I'd tend to try to put common functionality together where it makes sense -- having to hold context from line 9 to line 300 of a file can be a bit tricky for the best of us.

evankanderson (commented):
A few formatting issues, but the test cases look good and fairly comprehensive.  Given that we're covering at least 12 rules, 360 lines feels pretty compact, though I think we could spend a few lines on documentation as to the intent of the rules, and do a little more work to group the different "chunks" so that they aren't mixed together. Roughly the chunks are: `{enabled: True}` rules, `required_pull_request_reviews: {setting: value}`, and the top level "branch protection exists".

evankanderson (approved):
Approving; up to you if you want to address the remaining comments or not.

PR #435, opened 2026-09-09
Suffix GitLab security-baseline rule names with -gitlab

intelligent-ears:
Per discussion on mindersec/minder#6653.

`security-baseline/rule-types/gitlab/` and `security-baseline/rule-types/github/` declared identical `name:` values for all seven rules. Rule type lookup is keyed on project and name — `GetRuleTypeByNameParams` carries only `Projects` and `Name`, and the query doesn't filter on the `rule_type.provider` column — so the two sets can't coexist in a project, and a GitLab baseline profile would have no way to reference the GitLab versions.

Renaming these rather than the GitHub ones because they're `release_phase: alpha` and referenced by no profile, so there's no orphan problem: nobody has them applied, and `minder apply -f` has no way to know a renamed rule supersedes its predecessor.

`security-baseline-level-1.yaml` is `context.provider: github` and references the GitHub rules, so it's unchanged.

This is the going-forward convention @evankanderson proposed on #6653 — provider suffix on the variant, bare name on the original.


evankanderson (approved):
One minor correction: 

The `context.provider` field on Profiles (like the one on RuleTypes) is not used for anything, so its presence in the current baseline level 1 profile is not meaningful.

intelligent-ears:
Good to know — that also means removing it from the GitLab rules during #412's review wasn't what caused this; the collision was there from the moment those files landed. Is context.provider worth deprecating at some point, now that provider_traits covers the actual scoping? It reads like it does something.😅

evankanderson:
> Good to know — that also means removing it from the GitLab rules during #412's review wasn't what caused this; the collision was there from the moment those files landed. Is context.provider worth deprecating at some point, now that provider_traits covers the actual scoping? It reads like it does something.😅

I'm not sure what deprecating it would look like, but I think removing it from the examples in this repo makes sense.

In particular, we could maintain wire compatibility but break library compatibility by defining a "ProjectContext" proto that only contains the project and not the provider, but the current "Context" proto is shared across resources that do and don't require a provider.

PR #436, opened 2026-09-13
Split osps-ac-03-02 into GitHub and GitLab variants

### LINKED ISSUE #6653: osps-ac-03-02: use `provider_traits` to unblock GitLab deletion-protection semantics
Use the new `provider_traits` field to split or scope `osps-ac-03-02` (branch deletion protection) so GitHub and GitLab variants can diverge cleanly instead of trying to force one rule type to handle both APIs' differing semantics.

Depends on: sub-issues 1 and 2.

intelligent-ears:
Closes mindersec/minder#6653.

Adds `provider_traits: ["rest", "github"]` to `osps-ac-03-02` and introduces `osps-ac-03-02-gitlab` with `["rest", "gitlab"]`, following the naming convention established in #435.

### On the GitLab semantics

GitLab has no per-branch deletion setting. Protected branches can't be deleted via `git push` by anyone regardless of role, and deletion through the UI or API requires Maintainer — both are fixed platform behaviour rather than anything configurable, so there's no field corresponding to GitHub's `allow_deletions`. The protected-branch object exposes `allow_force_push`, access levels, and code-owner approval, and nothing else.

The GitLab rule therefore checks that the default branch is protected at all, which is what satisfies AC-03.02 on that platform.

Worth flagging: this is strictly weaker than `osps-ac-03-01-gitlab`, which checks `allow_force_push == false` on the same endpoint and also fails when the branch is unprotected. So `-03-02-gitlab` can't fail independently — it'll be green whenever `-03-01-gitlab` is. Keeping it anyway per @evankanderson, since the value is being able to attest the control is satisfied, the same way `osps-ac-02-01` exists on GitHub despite being trivially satisfied by the collaborator model.

No remediation, matching the other seven GitLab security-baseline rules — protecting a branch via the API requires choosing push, merge, and unprotect access levels, and there's no unambiguous "just prevent deletion" call to make.

### Notes

- No GitLab baseline profile exists yet, so this rule isn't wired into one. Same is true of the seven rules from #435.
- None of the existing GitLab security-baseline rules declare `provider_traits`; they predate the field and could use a follow-up pass.


evankanderson (approved):
Do you want to do a pass through the existing ruletypes and add `provider_traits` to most of them?

intelligent-ears:
Yes, happy to. Proposing to split it so it stays reviewable rather than one PR touching every rule type:

1. **`rule-types/common/`** — all `ingest.type: git`, so `["git"]`. Small and uncontroversial.
2. **`security-baseline/` (both trees)** — mostly `["rest", "github"]` / `["rest", "gitlab"]`, matching what just landed here.
3. **`rule-types/github/` and `rule-types/gitlab/`** — the largest, and the one that needs a decision first.

On 3: several rules under `rule-types/github/` ingest via `git` rather than `rest` — `actions_check_default_permissions`, `dependabot_configured`, `codeql_enabled`, `trivy_action_enabled`, `no_binaries_in_repo` and a few others. Mechanically they'd run fine against GitLab, since they just read repo contents. But they're about GitHub-specific subject matter, so they'd evaluate and produce meaningless results rather than being skipped.

So the question is whether those get `["git"]`, which is technically accurate about what they need, or `["git", "github"]`, which is accurate about where they make sense. The second uses `provider_traits` as a relevance gate rather than a capability gate, which may or may not be what the field is for.

Happy to start on 1 and 2 while we settle that — they're unambiguous either way.

evankanderson:
That's a good question on the `actions` (and `dependabot` / `codeql`) checks -- I can see a good argument for flagging those as `github` as well as `git`.  (`no_binaries_in_repo` _should_ be independent, and maybe should move to `common` -- I suspect there are a number of those, since we started writing rules for GitHub before we added any other provider support.)

I think we can handle those on a case-by-case basis during review (since I'll be looking at each rule anyway during the review).

intelligent-ears:
That works — case-by-case during review is easier than agreeing a rule in the abstract.

So `["git", "github"]` for the Actions/Dependabot/CodeQL ones, treating `github` as a relevance gate rather than only a capability one. That's the question answered.

On `no_binaries_in_repo` and friends: agreed, and I'll flag candidates for moving to `common/` as I go rather than quietly leaving them where they are. A directory move isn't a rule rename, so there's no orphan problem — the name stays the same.

Starting with `rule-types/common/` and the security-baseline trees since those are unambiguous, then `rule-types/github/` where the interesting calls are.

PR #437, opened 2026-09-19
Add provider_traits to common rule types

intelligent-ears:
First batch of the pass @evankanderson asked for in mindersec/minder-rules-and-profiles#436.

All seven rules in `rule-types/common/` get `provider_traits: ["git"]`. Five ingest via `git` directly; `osv_vulnerabilities` and `sonatype_oss_index_vulnerabilities` ingest via `deps`, which asserts the same `interfaces.GitProvider` in `internal/engine/ingester/ingester.go`, so `git` covers both.

The two vulnerability rules also use data sources (`osv`, `sonatype-oss-index`). Those call external APIs through the datasource machinery rather than the provider's REST client, so they don't add a trait requirement — worth noting since the same reasoning will apply to other datasource-using rules later in this pass. Data sources *can* take a provider for authenticated cases (`internal/datasources/service/options.go`), so a rule using an authenticated datasource might need more; neither of these does.

Next batches: `security-baseline/` both trees, then `rule-types/github/` and `rule-types/gitlab/` where the per-rule calls you mentioned come in.


evankanderson (approved):

evankanderson:
Thanks, these look good and seem universally applicable to source code!

PR #438, opened 2026-09-20
Add provider_traits to REST-based security-baseline rule types

intelligent-ears:
Second batch of the pass from #436, following #437.

Covers the REST-ingesting rules in both security-baseline trees — 10 GitHub (`["rest", "github"]`) and 7 GitLab (`["rest", "gitlab"]`). All seventeen call their provider's API directly, so these are mechanical.

### Held for a follow-up

The eleven `git`-ingesting rules in `security-baseline/rule-types/github/` need per-rule calls, so I've left them out to keep this diff boring. Preliminary sort:

**GitHub-specific in substance** — both glob `./.github/workflows/*.yml` and parse Actions syntax, so `["git", "github"]` per your note on #436:
- `osps-br-01-01` (untrusted input in CI)
- `osps-br-01-02` (branch name validation in CI)

**Provider-agnostic** — read repo contents with nothing GitHub about them, so `["git"]`, and arguably candidates for a `common/` tree in the security-baseline like the `no_binaries_in_repo` case you mentioned:
- `osps-br-03-02` (releases via encrypted channels)
- `osps-br-07-01` (sensitive data in repo)
- `osps-do-01-01` (user guides)
- `osps-do-04-01` (support scope documented)
- `osps-gv-03-01` (contribution process explained)
- `osps-le-03-01` (LICENSE/COPYING present)
- `osps-qa-02-01` (package management file present)
- `osps-qa-05-01` (no generated executables)
- `osps-qa-05-02` (no binary artifacts)

Happy to send that as traits-only, or as traits plus the moves, whichever you'd rather review.


evankanderson (approved):
I'm assuming that the `-LE-` baseline rules are outstanding because https://github.com/mindersec/minder/issues/6589 was awaiting a fix?

intelligent-ears:
No, nothing was held for #6589 — `osps-le-02-01`, `osps-le-02-02` and `osps-le-03-02` are all in this batch, since they ingest via `rest`.

The only `-LE-` rule outstanding is `osps-le-03-01`, and that's just because it reads LICENSE/COPYING off disk via `git`, so it landed in the provider-agnostic group with the other eight rather than being blocked on anything.

On batch 3 — did you have a preference between traits-only and traits plus moving the nine provider-agnostic ones into a `security-baseline/rule-types/common/` tree? Moves are free here since the rule names don't change, but it's a different review, so I'd rather ask than guess.

PR #440, opened 2026-09-22
Add provider_traits to the CI-parsing security-baseline rules

intelligent-ears:
Third batch from #436, following #437 and #438.

The two CI-parsing rules in the security-baseline GitHub tree get `["git", "github"]`. Both ingest via `git`, so `git` is what they mechanically require — but both glob `./.github/workflows/*.yml` and parse Actions syntax, so they'd evaluate against a GitLab repo and produce meaningless results rather than being skipped. `github` here is the relevance gate you described on #436 rather than a capability one.

Still outstanding: the nine provider-agnostic `git`-ingesting rules listed on #438. Those are `["git"]` on the merits, but they're also the `no_binaries_in_repo`-style candidates for moving to a `common/` tree, so I've left them until you've said whether you'd rather review traits-only or traits plus the moves.


evankanderson:
Let's try moving the common `git` rules into a `common/` directory alongside the provider_traits.  We'll learn something, as I think these are getting synced for at least a couple users.

evankanderson (approved):

PR #441, opened 2026-09-22
Add provider_traits to REST-based rule types

intelligent-ears:
Fourth batch from #436, following #437, #438 and #440.

All REST-ingesting rules under `rule-types/` — 26 GitHub (`["rest", "github"]`) and 2 GitLab (`["rest", "gitlab"]`). Mechanical; every one calls its provider's API directly.

### Remaining in `rule-types/`

- **4 `diff`** (`invisible_characters_check`, `mixed_scripts_check`, `pr_trusty_check`, `pr_vulnerability_check`) — `ingester.go` asserts `interfaces.GitHubListAndClone` for diff, so these need `github`. Want to check whether that interface implies `git` as well before declaring.
- **2 `artifact`** (`artifact_attestation_slsa`, `artifact_signature`) — the switch in `ingester.go` passes the provider straight to `NewArtifactDataIngest` without asserting, so the requirement lives inside the artifact ingester. Needs a closer look, and #6775 just changed that code.
- **22 `git`** — the judgement group, same shape as the nine still outstanding on #438. Includes `no_binaries_in_repo` (your move-to-common candidate), the Actions/Dependabot/CodeQL ones you ruled `["git", "github"]`, and two GitLab rules that presumably want `["git", "gitlab"]`.

I'll send the `git` ones once you've said which way you'd like the provider-agnostic rules handled — traits-only or traits plus moves. Happy to send `diff` and `artifact` separately either way.


evankanderson (approved):

evankanderson:
I'll look at the ones you called out later tonight, thanks!

evankanderson:
> Fourth batch from #436, following #437, #438 and #440.
> 
> All REST-ingesting rules under `rule-types/` — 26 GitHub (`["rest", "github"]`) and 2 GitLab (`["rest", "gitlab"]`). Mechanical; every one calls its provider's API directly.
> 
> ### Remaining in `rule-types/`
> * **4 `diff`** (`invisible_characters_check`, `mixed_scripts_check`, `pr_trusty_check`, `pr_vulnerability_check`) — `ingester.go` asserts `interfaces.GitHubListAndClone` for diff, so these need `github`. Want to check whether that interface implies `git` as well before declaring.

It would be good to deprecate the *built-in* implementation of these rules, and to upgrade these to a provider-independent `diff` API.  (Well, the `pr_trusty_check` can be removed -- the public Trusty API was shut down over a year ago.)

> * **2 `artifact`** (`artifact_attestation_slsa`, `artifact_signature`) — the switch in `ingester.go` passes the provider straight to `NewArtifactDataIngest` without asserting, so the requirement lives inside the artifact ingester. Needs a closer look, and #6775 just changed that code.

It would be desirable for these to work across OCI providers.  I'm not sure that they do today, but we should steer them in that direction.

> * **22 `git`** — the judgement group, same shape as the nine still outstanding on [Add provider_traits to REST-based security-baseline rule types #438](https://github.com/mindersec/minder-rules-and-profiles/pull/438). Includes `no_binaries_in_repo` (your move-to-common candidate), the Actions/Dependabot/CodeQL ones you ruled `["git", "github"]`, and two GitLab rules that presumably want `["git", "gitlab"]`.

I think the non-move-to-common rules got merged, but let's put the `git`-only rules (i.e. no provider-specific API dependencies) into `ruletypes/common`, which probably didn't exist when we started writing them.

Thanks for doing this!

PR #442, opened 2026-09-22
Add provider_traits to diff-based rule types

intelligent-ears:
Fifth batch from #436, following #437, #438, #440 and the REST batch.

The four `diff`-ingesting rules get `["git", "github"]`. Unlike the CI rules in #440, where `github` was a relevance gate, this is a genuine capability claim: `internal/engine/ingester/ingester.go` asserts `interfaces.GitHubListAndClone` for the diff ingester, and that interface requires both `ListFiles` (GitHub PR API) and `Clone` (git). Either missing and construction fails today.

### The two artifact rules

`artifact_signature` and `artifact_attestation_slsa` are the last of the non-`git` group, and they've hit a gap. Their real requirement is `provifv1.ArtifactProvider` — `artifact.go:127` asserts it to call `GetArtifactVersions`, and that's what fails if unsatisfied. The GitHub/OCI assertions further down are only for registry resolution and fall through to empty rather than erroring.

But `ArtifactProvider` has no entry in `providerTypeMap`, so no trait expresses it. It's a single-method interface that doesn't embed `Provider` or `ImageLister`, so `image-lister` would be claiming something untrue.

Options as I see them:

1. Leave them untraited — honest, since nothing in the current vocabulary describes what they need.
2. Declare `["github"]` as a relevance gate — true in practice, since GHCR is the only artifact provider wired up, but it doesn't describe the requirement.
3. Add an `artifact` trait to `ProviderType` with an `ArtifactProvider` entry in `providerTypeMap` — a minder-side change, and arguably right given the "requires API Y" framing on mindersec/minder#6650.

I'd lean towards 1 now and 3 as a follow-up if artifact rules ever need gating, but it's your call.


evankanderson (approved):
As I commented elsewhere, it would be great to be able to rework these four rules (well, three, since the Trusty rule is obsolete) to operate with PR traits that would work across GitHub and GitLab.  There's nothing intrinsic in what they look at that limits them to GitHub, it's simply an artifact of the current implementation.

intelligent-ears:
Agreed — nothing here is fundamentally GitHub-only, it's just what the current ingester/interfaces support. I'll open a follow-up issue to rework these (minus the obsolete Trusty one) around PR-level traits so they work across GitHub and GitLab, rather than trying to shoehorn it into this batch.

PR #444, opened 2026-09-23
Feat/provider traits git rules

intelligent-ears:
Final batch of the `provider_traits` pass from #436, following #437, #438, #440, #441 and #442. Does the moves alongside the traits, per your note on #440.

Two commits so the moves are readable separately from the trait additions.

### Stay in `rule-types/github/` — `["git", "github"]` (14)

All read `.github/`-specific paths or GitHub-only entity properties, so they'd evaluate to nothing meaningful elsewhere:

`actions_check_default_permissions`, `actions_check_pinned_tags`, `codeql_enabled`, `dependabot_configured`, `golangci-lint_github_action`, `grype_github_action_scan_container_image`, `grype_github_action_scan_repo`, `repo_action_allow_list`, `scorecard_enabled`, `trivy_action_enabled`, `trufflehog_github_action`, `workflow_no_pull_request_target`, `workflow_pull_request`, `source_code_is_public`

`source_code_is_public` is the non-obvious one — it ingests via `git` but never reads a file; it evaluates `input.properties.is_private`, which is a GitHub property name, so it isn't portable as written.

### Moved to `rule-types/common/` — `["git"]` (6)

`no_binaries_in_repo`, `openssf_bestpractices`, `security_insights`, `security_insights_dep_policy`, `security_policy`, `talisman_secrets_scanning`

All read repo contents only. `talisman_secrets_scanning`'s `_test.star` and `.testdata/` moved with it. Rule names are unchanged, so no orphan problem — the sync question you raised is purely about the file paths moving.

### `rule-types/gitlab/` — `["git", "gitlab"]` (2)

`gitlab_dependency_scanning_enabled`, `gitlab_pipeline_secret_detection_enabled`

### Still outstanding

The two `artifact` rules, per the trait-vocabulary gap on #442. And the nine provider-agnostic security-baseline rules — same move-to-common question, but that tree has no `common/` directory yet, so worth confirming you want one created there too.


evankanderson:
Oof, this will have conflicts with my #445 -- which do you want to merge first?

evankanderson (approved):

intelligent-ears:
You can merge #445 first — I'll rebase and update this one tomorrow. Have a quiz today so won't get to it right away.

PR #445, opened 2026-09-23
Package support files in txtar, adjust tests

evankanderson:
Packages all the existing supporting filesystem (and HTTP response) content in \[txtar format](https://pkg.go.dev/golang.org/x/tools/txtar).  This prevents false-positives from other scanning tools (like CVE detection in \_example\_ package.json files, or setting up Dependabot for the same).

I also converted all the `security-baseline/` tests into Starlark, since:

1. That is required for using the `txtar` format.
2. We removed the old ruletype testing framework without converting those tests, so they were effectively un-covered

For the `security-baseline` tests, I ended up coalescing all the tests for a given control area (Governance, Documentation, etc) into a single test file, to reduce the total number of files to manage.  I expect to be adding more test coverage and converting some of the rules to `.rego` format shortly as well.


security-baseline/rule-types/github/osps-br.star
```diff
@@ -0,0 +1,81 @@
...
+    keep = lambda file: file.find("ref") != -1
+    res = eval(
+        rule="osps-br-01-01",
+        entity=ENTITY,
+        mock_fs={k: workflow_files[k] for k in workflow_files if keep(k)}
+    )
+    assert.eq(res["status"], "fail")
+    assert.eq(res["message"],
+    '''evaluation failure: Evaluation failures: 
+ - Workflow .github\\workflows\\pr_test_ref.yaml has a dangerous trigger and checks out a pull request in job 'test'
```
krrish175-byte: The old test YAML had forward slashes in the workflow paths (e.g. `.github/workflows/pr_test.yaml`), but this assertion expects double backslashes. Since the tests are passing in CI (Linux), the rule must actually be producing backslash-separated paths when given txtar-sourced mock_fs data. That makes sense if the mock_fs layer uses OS-native separators, but it means these tests might behave differently in production where real repo paths come through with forward slashes. Worth double-checking that the path format coming from the real ingester matches what's being tested here.

security-baseline/rule-types/github/osps-br.star
```diff
@@ -0,0 +1,81 @@
...
+
+def test_br_01_01_unsafe_checkout():
+    keep = lambda file: file.find("ref") != -1
+    res = eval(
+        rule="osps-br-01-01",
+        entity=ENTITY,
+        mock_fs={k: workflow_files[k] for k in workflow_files if keep(k)}
+    )
+    assert.eq(res["status"], "fail")
+    assert.eq(res["message"],
```
krrish175-byte: The full multi-line error message assertion is quite brittle. If the error formatting changes (even something like adding a period or adjusting the "Multiple issues:" framing), this test will break for reasons unrelated to security. Checking for the presence of the workflow filename would catch real regressions without being as tightly coupled to the output format -- something like `assert.true(res["message"].find("pr_test_ref.yaml") != -1)` would be enough.

security-baseline/rule-types/github/osps-br.star
```diff
@@ -0,0 +1,81 @@
...
+
+def test_br_03_01_not_http():
+    # This shouldn't be true on GitHub, but if the repo URL were somehow HTTP...
+    res = eval(
+        rule="osps-br-03-01",
+        entity=ENTITY,
+        mock_http={
+            REPO_URL: body('{"clone_url": "http://github.com/mindersec/minder.git"}')
+        }
+    )
```
krrish175-byte: Missing newline at end of file. Same issue in osps-ac.star and osps-qa.star.

security-baseline/rule-types/github/osps-gv.star
```diff
@@ -0,0 +1,74 @@
...
+    )
+    assert.eq(res["status"], "fail")
+
+
+def test_gv_03_01_plain_file():
+    keep = lambda file: file == "CONTRIBUTING"
+    res = eval(
+        rule="osps-gv-03-01",
+        entity=ENTITY,
+        mock_fs={k:contributing[k] for k in contributing if keep(k)}
```
krrish175-byte: The `keep = lambda / dict comprehension` pattern shows up 4 times here and 7 times in osps-qa.star. A small helper at the top of each file would clean this up a lot:

```python
def pick(txtar, predicate):
    return {k: txtar[k] for k in txtar if predicate(k)}
```

Then the call sites become `mock_fs=pick(contributing, lambda f: f == "CONTRIBUTING")`. Not a blocker, but given how many more tests are coming, better to establish the pattern now.

krrish175-byte (commented):
The txtar consolidation is a great improvement -- the old test directories full of special-looking files like pyproject.toml and Gemfile.lock were definitely triggering Trivy. The Starlark conversion for security-baseline looks solid too, nice to finally have those covered. A few things worth a look below.

rule-types/common/enforce_file_test.star
intelligent-ears: This one swaps its fixture for an inline `"Test content"` rather than moving to txtar. Fine by me for a rule that only checks presence — just noting it's the one file that went a different way, in case that wasn't deliberate.

rule-types/common/license_test.star
intelligent-ears: Both cases now feed `files["MIT-LICENSE"]`, so the "doesn't match" test is really "MIT isn't Apache-2.0" rather than a genuinely different license. From the rename mapping, `license.testdata/license_doesnt_match/LICENSE` became `testdata/license.txtar`, and the txtar only carries an `MIT-LICENSE` entry — looks like the non-matching fixture got absorbed rather than packed.

It still fails for a real reason, so not a false green, but the coverage is narrower than the test name implies. Either a second txtar entry with a different license, or rename the test to what it now asserts?

rule-types/common/python_lock_file_exists_test.star
intelligent-ears: The poetry case (expects pass) and the no-lock case (expects fail) both now use `files["codegate/pyproject.toml"]`, where previously they pulled different pyproject files from `poetry_lock/` and `no_lock/`.

If the rule keys purely off the presence of `poetry.lock` this is fine and arguably cleaner. If it also inspects pyproject contents, the no-lock case may now fail for a different reason than intended. I couldn't tell from the diff alone.

intelligent-ears (commented):
The txtar consolidation is clearly right — a packed fixture per concern reads far better than the directory sprawl, and getting example `package.json`/`go.mod` files out of scanner range is worth it on its own. Converting the `security-baseline/` tests is the bigger win though; I hadn't registered that those had been left uncovered when the old framework went away.

Coalescing by control area seems fine. The test names stay specific enough that a failure still points somewhere useful.

@krrish175-byte's backslash finding is a good one — worth settling before merge, since a path-format mismatch between `mock_fs` and the real ingester means these tests pin something production doesn't produce.

My notes below are a different set, mostly fixtures that seem to have lost a distinction during the packing. Only the license one would I want changed before merge.

security-baseline/rule-types/github/osps-br.star
evankanderson: Oop, good point!  Let me see if forward slashes work.

rule-types/common/enforce_file_test.star
evankanderson: Yes, I did that because it was a one-line file, and there wasn't any other file content.  It didn't seem worth setting up the rest of the fixture for that.

rule-types/common/license_test.star
evankanderson: I can rename the test -- the deleted file was actually _also_ a copy of the MIT license, but the old framework needed two different files for two different tests.

rule-types/common/python_lock_file_exists_test.star
evankanderson: Both tests actually pulled the same `pyproject.toml` -- there were two copies of the file, in two different directories, but with the same contents.  One of the advantages of handling some of this in starlark is being able to trim / duplicate the filesystem without needing duplicate files.

security-baseline/rule-types/github/osps-br.star
evankanderson: Okay -- this is interesting (and may end up having to tie into your next point).  It looks like the slashes come from `testkit.WithGitFiles`, which creates a `billy.Filesystem` with the given paths, creating directories as it goes.  Given that implementation, it seems like switching to use linux-style paths in the filesystem would require rewiring the various OS modules.  Given that constraint, I'll adopt the next suggestion of matching substrings in the message.

security-baseline/rule-types/github/osps-br.star
evankanderson: The intent of the matching was to show the actual message (and its current doubling of content).  The content doubling seems to be something from the `pkg/ruletest` module, though, so I'll switch this to some substring matches to avoid the path problem earlier.

evankanderson: https://github.com/mindersec/minder/pull/6806 for the doubling.

security-baseline/rule-types/github/osps-gv.star
evankanderson: Since there was only one fileset per file (and we don't have a good library mechanism that I know of), I've implemented this as `filter_$type(picker)`.  I expect once we have a library/module system to create a central `filter` function.

evankanderson:
Thanks for the review!  I think I've addressed the concerns, and I appreciate the double-checking on removed test fixtures (which don't show up nicely in the diff output, unfortunately!).  Removing the duplication between the different files seems to have won around 1500 lines fewer of test files, so yay for that!

evankanderson:
Oh, and the `mindev test` suite is _so_ much nicer to work with and iterate on compared with the old Go framework.  It's 9s for the suite to run on slow GitHub runners, and almost instant on my desktop when idle.

